PR #5, opened 2020-08-17
[AA-943] Adjustments after bulk copy from old repo

plioi:
1. Consistently install Microsoft.NET.Test.Sdk 16.4.0 across test projects in order to account for missing automatic binding redirect generation. All changes in this first commit were automated by NuGet during installation of that package. This addresses situations where tests will pass on one developer's machine or another inconsistently. It has been reported as a problem with the Moq testing library but is in fact an issue with NetStandard 2.0 in combination with .NET Framework: https://github.com/dotnet/standard/issues/481 Note that the suggested msbuild tags for test projects in that referenced Issue are in fact handled by referencing the Test SDK.
2. Remove .gitignore rule for `.../WebRoot/post-install-readme.txt` as this file is no longer produced during builds.
3. The TeamCity readme file suggests the current default branch, `main`, as an appropriate setting for the git.branch.default TeamCity parameter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,5 +77,4 @@ EdFi.Suite3.Installer.AdminApp/downloads
 EdFi.Suite3.Installer.AdminApp/Ed-Fi-Common
 EdFi.Suite3.Installer.AdminApp/Ed-Fi-ODS
 EdFi.Suite3.Installer.AdminApp/Ed-Fi-ODS-Implementation
-EdFi.Suite3.Suite3.Ods.AdminApp/WebRoot/post-install-readme.txt
 
--- a/.teamcity/readme.md
+++ b/.teamcity/readme.md
@@ -12,7 +12,7 @@ version: `2019.2.2`):
     github.username = <username>
     github.accessToken.protected (password type) = <access token>
     github.accessToken = %github.accessToken.protected%
-    git.branch.default = development-v3
+    git.branch.default = main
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -77,5 +77,4 @@ EdFi.Suite3.Installer.AdminApp/downloads
 EdFi.Suite3.Installer.AdminApp/Ed-Fi-Common
 EdFi.Suite3.Installer.AdminApp/Ed-Fi-ODS
 EdFi.Suite3.Installer.AdminApp/Ed-Fi-ODS-Implementation
-EdFi.Suite3.Suite3.Ods.AdminApp/WebRoot/post-install-readme.txt
```
stephenfuqua, line 80: Just two week ago I saw this file being created when running `build.ps1 package`.

.teamcity/readme.md
```diff
@@ -12,7 +12,7 @@ version: `2019.2.2`):
     github.username = <username>
     github.accessToken.protected (password type) = <access token>
     github.accessToken = %github.accessToken.protected%
-    git.branch.default = development-v3
+    git.branch.default = main
```
stephenfuqua, line 15: Good catch.

.gitignore
stephenfuqua: You're right, its no longer present.

PR #6, opened 2020-08-17
[AA-944] Favor OnPremisesRelease Configuration

plioi:
To meet the behavior of TeamCity builds prior to the creation of the Ed-Fi-Alliance-OSS organization, the TeamCity build favors the OnPremisesRelease configuration.

Note also some adjustments to the build.ps1 script to better reflect the current Build Configurations matrix in the *.sln file itself, to ensure all the intended tests can be found and run.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -102,6 +102,9 @@ $solution = "Application\Ed-Fi-ODS-Tools.sln"
 if ("Release" -eq $BuildConfiguration) {
     $configuration = "Release"
     $testConfiguration = "Release"
+} elseif ("OnPremisesRelease" -eq $BuildConfiguration) {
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -100,10 +100,12 @@ param(
 $solution = "Application\Ed-Fi-ODS-Tools.sln"
 
 if ("Release" -eq $BuildConfiguration) {
+    $configuration = "Release"
+    $testConfiguration = "Release"
+} elseif ("OnPremisesRelease" -eq $BuildConfiguration) {
```
stephenfuqua, line 105: You'll need to modify line 75 above.

plioi: Yes, thanks. I'll push that shortly.

PR #13, opened 2020-08-25
[AA-1013] - Bulk Upload works with an unrelated API key and secret

saa14:
**Description**
- Added a new SaveBulkUploadCredentialsModel with only ApiKey and ApiSecret properties. Added the SaveBulkUploadCredentialsModelValidator and validation rules for the ApiKey and ApiSecret properties.
- Refactored to use SaveBulkUploadCredentialsModel instead of BulkFileUploadModel in the BulkUploadValidationTests, WhenRunningBulkUpload and SaveBulkLoadCredentials.
- Added a partial view SaveBulkUploadCredentialsForm by extracting the form from the BulkLoad view.
- Added BulkUploadValidationTests to the tests project in order to test the new validation rules added to the SaveBulkUploadCredentialsModel.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Management.Tests/Database/Commands/BulkUploadValidationTests.cs
+++ b/Application/EdFi.Ods.AdminApp.Management.Tests/Database/Commands/BulkUploadValidationTests.cs
@@ -41,15 +41,20 @@ namespace EdFi.Ods.AdminApp.Management.Tests.Database.Commands
 
             SetupApplicationForInstance(odsInstance2);
 
-            var bulkUploadModel = new BulkFileUploadModel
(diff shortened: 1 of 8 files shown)


plioi:
@saa14 Please scan my updates to see if you agree with them. I wanted to have a single round-trip to the database for the EF query, I wanted to be as pessimistic about nulls as possible, and I saw ReSharper suggesting I collapse it down using the `?.` operator once I had all the possible null checks written explicitly.

The tests still pass, and this worked in my local testing in Year Specific mode. It correctly rejected a typo, correctly rejected a real key/secret for the wrong instance, and it correctly accepted a real key/secret for the right instance.

plioi:
@saa14 Verified my changes made during review, and confirmed that his testing included single instance mode.

PR #23, opened 2020-09-17
[AA-1704] Updated the TeamCity version parameter from 2.0.0 to 2.1.0 for Admin App

saa14:
**Description**

Updated the TeamCity version parameter from 2.0.0 to 2.1.0 for Admin App.


plioi (approved):
We witnessed the expected artifact number after pushing this PR up.

PR #32, opened 2020-09-29
[AA-1050] preemptively resolve simple compilation errors that would eventually surface in ASPNET Core project

plioi:
This is meant to be a low-risk PR that improves the ASPNET Core project's potential to compile once downstream dependencies are compiling for netcoreapp3.1.

I uncommented the inclusion of all cs files in the project, provoking the many expected compiler errors, then I installed several NuGet packages at modern version numbers, based on those error messages, reducing the number of errors from ~1430 to ~291 (exact numbers can be misleading when there are so many). Then, I commented out the inclusion of the problematic cs files again, because we'll need our dependencies truly targeting netcoreapp3.1 before these can be expected to work.

The changes within EdFi.Ods.AdminApp.Web.Core.csproj are extremely low risk because this project is still not the one we'd deploy in the short term.

The changes within EdFi.Ods.AdminApp.Web/* are INTENDED to be low risk in the following way, so the reviewer should pay careful attention to these file edits since this is where a typo could allow for a (very unlikely) behavior change to slip in:

1. In the legacy web app csproj, I define NET48. This is automatic in modern projects, but had to be explicit here. This allows us to say #if NET48 in our shared c# files and have that be treated as "true" in the legacy project and "false" in the ASPNET Core project.
2. For several "using" directives, continue to include the same existing namespace #if NET48 so that there are no changes to the deployable code right now.
3. When **not** NET48, use slightly different namespaces on new lines.

In other words, all the diffs in cs files here should be essentially "no change" when NET48, and "trivial alternative" when !NET48.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web.Core/EdFi.Ods.AdminApp.Web.Core.csproj
+++ b/Application/EdFi.Ods.AdminApp.Web.Core/EdFi.Ods.AdminApp.Web.Core.csproj
@@ -64,6 +64,7 @@
     <PackageReference Include="EdFi.Admin.LearningStandards.Core" Version="1.1.0" />
     <PackageReference Include="FluentValidation" Version="9.2.2" />
     <PackageReference Include="FluentValidation.AspNetCore" Version="9.2.0" />
+    <PackageReference Include="log4net" Version="2.0.11" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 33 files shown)


plioi:
@saa14 The build had failed, and I have corrected that error. This is ready to review when convenient.

saa14 (approved):

PR #43, opened 2020-10-08
[AA-1104] Wire up FluentValidation in Startup

CSR2017:
@plioi @saa14 

- Registered fluent validation on configure services on startup

- Added temporary view, model, and validator for testing.


plioi:
Thanks. The sample code was valuable proof, so we know we won't run into trouble later when we start standing up each "feature vertical". This made it very easy to test. As agreed, I'm reverting that commit so that it won't merge, but it will still be in history for reference later if we want.

plioi (approved):

PR #53, opened 2020-10-21
[AA-1128] - Wire up Hangfire and Signalr in Startup for .NET core project

saa14:
**Description**

- Added SignalR client library to the .net core project.
- Added temporary files to create a SignalR hub and added SingalR client code to demo a proof-of-concept chathub and client for SignalR.
- Added Hangfire.SqlServer, Hangfire.Core and Hangfire.AspNetCore packages to the .net core project.
- Configured Hangfire in the .net core project Startup. 
- Added temporary logic to test that Hangfire works. Used Hangifre.SqlServer. **This test needs an empty "HangfireTest" database and the connection string pointing to the server.**


plioi:
- [ ] Resolve merge conflicts introduced by recent items merged to `main`. It is probably a trivial conflict in Startup.cs.
- [ ] Study PR #51. The new JS files you're adding have a better home now under wwwroot/Scripts. However, you will find we have a coincidentally named file "signalr.js" there already. I say it's a coincidence because it is not the old signalr library, but instead our related custom progress bar code. I think if you rename our progress bar file appropriately (take care in the Layout and bundleconfig.json for references to it), commit, and then move this real library signalr file down into that folder, that combination should work. Be sure to include signalr.min.js in the Layout near the footer where we now include all js libraries.
- [ ] I see that the 3 nuget packages you added in the csproj are consistent with the version we set up in a previous ticket. Note, though, that the already-referenced Hangfire 1.7.14 *already* includes Hangfire.AspNetCore, Hangfire.Core, and Hangfire.SqlServer of the same version. You can see this when you expand the Hangfire node inside Solution Explorer and in the "Dependencies" section here: https://www.nuget.org/packages/HangFire/1.7.14 so I think you could remove these 3 new PackageReference tags and you would find everything still works and that expanding Hangfire in Solution Explorer still shows the related packages do in fact arrive.

plioi (approved):

plioi:
I see how your proof-of-concept commit/revert showed setting up the Dashboard and connecting to SQL Server. I agree with leaving that out for now. I've updated AA-1118 with guidance on introducing it with respect to our SQL Server / Postgres needs and with respect to the older .NET 4.8 code we might need to phase out at the same time. It makes sense to do that while working on the Learning Standards feature since we'll have real feedback at that time.

PR #57, opened 2020-10-27
[AA-1106] Finalize Home Index for .NET Core

CSR2017:
@plioi @saa14 
Thanks


plioi:
At this point, I think the HomeController should be shared between the Web and Web.Core projects since they are (nearly) identical here. That would mean (*not* in Solution Explorer!), cutting the new HomeController from Web.Core and pasting it on top of the old one in Web, then using our csproj file sharing to make it available here in Web.Core as well. This would make it much more clear what the edge cases are. In this PR, it just *looks* like a whole new controller.

Application/EdFi.Ods.AdminApp.Web/Infrastructure/CloudOdsAdminAppSettings.cs
```diff
@@ -15,30 +15,42 @@ public class CloudOdsAdminAppSettings
     {
         private static readonly Lazy<CloudOdsAdminAppSettings> _instance =
             new Lazy<CloudOdsAdminAppSettings>(() => new CloudOdsAdminAppSettings());
-        private readonly AppSettings _appSettings = ConfigurationHelper.GetAppSettings();
+
+        private AppSettings AppSettings
```
plioi, line 19: I like how this lets us wait until it's easier to phase out the old concepts. Great idea!

Application/EdFi.Ods.AdminApp.Web.Core/Controllers/HomeController.cs
```diff
@@ -4,20 +4,65 @@
...
+        {
+            _cloudOdsUpdateService = cloudOdsUpdateService;
+            _homeScreenDisplayService = homeScreenDisplayService;
+            _getOdsInstanceRegistrationsQuery = getOdsInstanceRegistrationsQuery;
+        }
+
+        public async Task<ActionResult> Index(bool setupCompleted = false)
         {
-            return View();
+        #if NET48
```
plioi, line 40: Let's talk about how we can avoid the need  for #if here.

We need to only introduce #if NET48 for code we intend to *ruthlessly delete without thinking* later. Since this is code we *need*, this use of #if makes it *look* like dead code we can throw away soon.

Application/EdFi.Ods.AdminApp.Web/Display/HomeScreen/BaseHomeScreenDisplayService.cs
```diff
@@ -52,10 +52,12 @@ public virtual List<HomeScreenDisplay<HomeScreenEnumeration>> GetHomeScreenDispl
                     x.HomeScreen == HomeScreenEnumeration.Settings).IsEnabled = false;
             }
 
+            #if NET48
```
plioi, line 55: Same concern as above. This makes it look like we are *supposed* to delete this code forever soon.

Application/EdFi.Ods.AdminApp.Web.Core/Views/Home/Index.cshtml
```diff
@@ -26,7 +26,8 @@ See the LICENSE and NOTICES files in the project root for more information.
     var UpdatesDisplay = Model.HomeScreenDisplays.FirstOrDefault(a => a.HomeScreen == HomeScreenEnumeration.Updates);
 }
 
-@if (Request.IsAuthenticated)
+@* #if NET 48 @if (Request.IsAuthenticated)*@
```
plioi, line 29: Because I'm hitting roadblocks on logins, this is going to be a continued problem. We need a solution that will make it very clear to us later what the right final answer is, and this would be very easy to miss at that time.

I think we need to instead extract a helper extension method so this line can be something like Request.IsAuthenticatedOrTemporarilyAssumingAuthenticated() or something, and the implementation for NET48 would be to return Request.IsAuthenticated while for !NET48 it would return true with a very, very clear comment that it is a temporary measure waiting on AA-1120. Then we can't forget to get it right in time.

plioi:
Be sure to investigate the build failure.

plioi (approved):

plioi:
I made a few additions in order to get the tests to pass. When the tests run under .NET Core we needed to take a few extra steps to ensure that we were getting the new config value infrastructure initialized just like during a run of the web application. This is great, because it motivated some setup we'd like to start adopting more and more in the tests, so the first step of that appears as part of this PR.

PR #75, opened 2020-11-13
[AA-1140] Build .NET Core deployment package without affecting .NET Framework deployment package

CSR2017:
@plioi @saa14 
Please help review and merge changes.
Thanks

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -342,9 +342,15 @@ function Invoke-Build {
  
     if($netCore)
     {
-        $outputPath = "$PSScriptRoot/Application/EdFi.Ods.AdminApp.Web.Core/publish"        
-        $project = "$PSScriptRoot/Application/EdFi.Ods.AdminApp.Web.Core/EdFi.Ods.AdminApp.Web.Core.csproj" 
-        dotnet publish $project -c $configuration /p:EnvironmentName=$configuration -o $outputPath
+        $baseDir = "$PSScriptRoot/Application"
+        $projectPaths = Get-ChildItem -Path $baseDir -Include *.csproj -Recurse -exclude EdFi.Ods.AdminApp.Web.csproj,EdFi.Ods.AdminApp.Management.Tests.csproj       
+        foreach ($projectPath in $projectPaths) {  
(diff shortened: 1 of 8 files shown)


plioi:
image

`CreateDefaultBuilder` in `Main` already sets up our `Configuration` object, and more completely than we see in this PR. Note the tool tip includes, among many other things set up, appsettings.json and appsettings.ENVIRONMENTNAME.json support. Since this set up so much for us by default, and we are now throwing that out in Startup, I think we're missing something important.

I have been able to successfully use appsettings.json and appsettings.OnPremises.json already, for example. This would also quietly lose the default support for loading environment variables into `IOptions<AppSettings>` which will be useful later when we do Docker support.

If the defaults aren't enough (but should be for json files already), you can more safely affect the Configuration over in main rather than building it from scratch in Startup.

plioi:
Let's also talk about how much we can follow the pattern set up here by Stephen for use by TeamCity. It might be time to make .NET Core builds the official output of TeamCity now, but it deserves discussion first.

build.ps1
```diff
@@ -332,14 +339,23 @@ function PushPackage {
...
-    Invoke-Step { InitializeNuGet }
-    Invoke-Step { Clean }
-    Invoke-Step { Restore }
-    Invoke-Step { AssemblyInfo }
-    Invoke-Step { Compile }
-    Invoke-Step { RevertAssemblyInfo }
+ 
+    if($netCore)
+    {
+        $outputPath = "$PSScriptRoot/Application/EdFi.Ods.AdminApp.Web.Core/publish"
```
plioi, line 345: I understand that the dotnet publish command will also do a restore and build if necessary, but I think it will only do so for the specified project (Web.Core) and its dependencies. So, I bet that this does not build the whole solution... it probably leaves out the building of the tests, so the step that runs tests might be doing less than we think it is.

plioi: So I think we might want to do a dotnet build here solution-wide, and then use dotnet publish later where we need to prepare the package. dotnet publish has a --no-build option to avoid re-building in that case.

plioi (approved):

PR #76, opened 2020-11-14
AA-1108 Fix error saving application without application name

leonardoSotoGap:
Add condition to IsSystemReservedApplicationName method in Application Extension file to do not take null values.
It Fixes Adding and Editing an Application.


plioi:
Leo's commit signature was invalid due to email mismatch. I reviewed the code enough to feel confident signing it myself in his absence. He still appears as the true Author, I as the Committer. I'd like a separate person to therefore be the official reviewer, tester, and merger of the PR.

saa14 (approved):

PR #78, opened 2020-11-17
[AA-1151] - Phase out DatabaseProviderHelper usages on .NET Core builds

saa14:
**Description**

- Refactored UpgradeEngineFactory to use the DatabaseEngine appsetting instead of DatabaseProviderHelper for the .NET Core builds.
- Refactored DatabaseConnectionProvider to use the DatabaseEngine appsetting instead of DatabaseProviderHelper for the .NET Core builds.
- Refactored ReportsConfigProvider to use the DatabaseEngine appsetting instead of DatabaseProviderHelper for the .NET Core builds.


plioi:
@saa14 Review the build failure. One test is affected by the constructor change in UpgradeEngineFactory. You may need to stub an IOptions conditionally here. We had to do similar over in RegisterOdsInstanceCommandTests so take a look there in case that provides a quick suggestion.

plioi (approved):

PR #79, opened 2020-11-18
[AA-1153] - Fix problem when browsing to Reports when using non default configuration

leonardoSotoGap:
Use another way to navigate between pages using razor html helpers with parameters


saa14 (changes requested):
Although, the change works for the .NET48 project, please make the same change for the _SelectDistrict.cshtml view in the .NET Core project as well.

saa14 (approved):

PR #83, opened 2020-11-25
Correct the teamcity-get-package-version script to seek the right package name.

plioi:
Now that there is a single package, remove "Core" from the expected name as it is no longer used to distinguish old from new packages.


andonyns (approved):

jasonh-edfi (approved):
Seeing the need to remove "Core" from the build scripts

PR #84, opened 2020-11-25
AA-1157 .NET Core global exception handling

plioi:
This fixes global exception handling so that all uncaught exceptions can be logged. See the individual commit messages for the explanation of each step.


leonardoSotoGap (approved):
Great work.

PR #85, opened 2020-11-26
[AA-1067] Handle repeated OdsInstances in Bulk load

leonardoSotoGap:
Handle repeated OdsInstances in Bulk load.
For one side we are now skipping data rows with NumberSuffixes that were Registered Previosly in Validations because this is not exactly an error.
For the  other side we are now controlling that the instances that are going to be created must be only the new ones.

I also Added integration tests to BulkRegisterOdsInstancesCommand class.


Application/EdFi.Ods.AdminApp.Management/Instances/BulkRegisterOdsInstancesCommand.cs
```diff
@@ -24,16 +25,33 @@ public BulkRegisterOdsInstancesCommand(RegisterOdsInstanceCommand registerOdsIns
         public async Task<IEnumerable<BulkRegisterOdsInstancesResult>> Execute(IEnumerable<IRegisterOdsInstanceModel> odsInstances, ApiMode mode, string userId, CloudOdsClaimSet cloudOdsClaimSet = null)
         {
             var results = new List<BulkRegisterOdsInstancesResult>();
-            foreach (var instance in odsInstances)
+
+            var odsInstancesRecords = odsInstances;
```
plioi, line 29: Do we need two variables for the same object here?

Application/EdFi.Ods.AdminApp.Management/Instances/BulkRegisterOdsInstancesCommand.cs
```diff
@@ -24,16 +25,33 @@ public BulkRegisterOdsInstancesCommand(RegisterOdsInstanceCommand registerOdsIns
...
+            var newOdsInstanceToRegister = _registerOdsInstanceCommand.PreExistingOdsInstanceRegistrations(odsInstancesRecords, mode);
```
plioi, line 30: The variable and method naming here disagree with each other. The variable calls these the new ones to register. The method calls it the preexisting ones that we would NOT register. Which is the truth?

Application/EdFi.Ods.AdminApp.Management/Instances/RegisterOdsInstanceCommand.cs
```diff
@@ -16,15 +18,15 @@ public class RegisterOdsInstanceCommand
     {
         private readonly IOdsInstanceFirstTimeSetupService _odsInstanceFirstTimeSetupService;
         private readonly IDatabaseConnectionProvider _connectionProvider;
-        private readonly AdminAppIdentityDbContext _identity;
+        private readonly AdminAppIdentityDbContext _identity;
```
plioi, line 21: Let's avoid checking in whitespace-only changes like this.

Application/EdFi.Ods.AdminApp.Management/OdsInstanceServices/OdsInstanceFirstTimeSetupService.cs
```diff
@@ -66,5 +67,10 @@ private async Task AddOdsInstanceRegistration(OdsInstanceRegistration odsInstanc
             _database.OdsInstanceRegistrations.Add(odsInstanceRegistration);
             await _database.SaveChangesAsync();
         }
+
+        public IEnumerable<OdsInstanceRegistration> PreExistingOdsInstanceRegistrations()
+        {
+            return _database.OdsInstanceRegistrations;
```
plioi, line 73: Returning a whole DbSet as an IEnumerable can cause some surprises, like unintented "SELECT *" from a large table. If we remove this method and just use "_database.OdsInstanceRegistrations" where this method was called, it might be more clear whether or not we are getting efficient queries.

plioi:
@leonardoSotoGap I've added some feedback comments. Also, please update this PR title with the JIRA number like we do on other PRs, and merge from "main" after you are ready for another round of review so that the PR is as up to date as possible.

Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/OdsInstances/RegisterOdsInstanceModel.cs
```diff
@@ -127,5 +128,11 @@ private static bool BeAUniqueInstanceDescription(string newInstanceDescription)
 
             return !_database.OdsInstanceRegistrations.Any(x => x.Description == trim);
         }
+
+        public IEnumerable<RegisterOdsInstanceModel> GetOdsInstancesToRegister(IList<RegisterOdsInstanceModel> dataRecords) {
+            var previousRegisters = _database.OdsInstanceRegistrations;
+            var newRows = dataRecords.Where(p => !previousRegisters.Any(p2 => p2.Name == InferInstanceDatabaseName(p.NumericSuffix)));
```
plioi, line 134: On this and similar lines in this PR, can we have more descriptive names than "p" and "p2"?

Application/EdFi.Ods.AdminApp.Management/Instances/BulkRegisterOdsInstancesCommand.cs
leonardoSotoGap: You are right, it was not necessary to create a copy for that variable. i am going to use just the original one.

Application/EdFi.Ods.AdminApp.Management/Instances/BulkRegisterOdsInstancesCommand.cs
leonardoSotoGap: You are right, i already changed the method to another more descriptive name.

Application/EdFi.Ods.AdminApp.Management/Instances/RegisterOdsInstanceCommand.cs
leonardoSotoGap: I forgot to Format this file(Control+k+d) to remove unnecessary blank spaces. But now it is.

Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/OdsInstances/RegisterOdsInstanceModel.cs
leonardoSotoGap: I already reviewed all files again to ensure changing those weird variables names by another ones more descriptive. Thanks

Application/EdFi.Ods.AdminApp.Management/OdsInstanceServices/OdsInstanceFirstTimeSetupService.cs
leonardoSotoGap: As we spoke in the call , i already included the .ToList() statement in the line that this method is called, to improve the performance. For that reason this method to get  PreExistingOdsInstanceRegistrations will continue

CSR2017: @leonardoSotoGap 
I am going through the changes on Command files. Wondering why we take the method all the way to OdsInstanceFirstTimeSetupService class. I do see the only reason is, we are having the db context there. I guess, we can have some simple common service which will be called from the model to provide filtered list, so that we can avoid two code flow (one from model and one from command class for achieving the same list)! We may hold the filtered list on FilteredDataRecords property on the BulkRegisterOdsInstancesModel ! This way it will be easy to maintain the filtration code. 
@plioi Please provide your insight. Thank you

CSR2017: > Returning a whole DbSet as an IEnumerable can cause some surprises, like unintented "SELECT *" from a large table. If we remove this method and just use "_database.OdsInstanceRegistrations" where this method was called, it might be more clear whether or not we are getting efficient queries.

I agree with Patrick's suggession on use _database.OdsInstanceRegistrations on calling place. Just add constructor dependency for  AdminAppDbContext.

Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/OdsInstances/BulkRegisterOdsInstancesModel.cs
```diff
@@ -90,7 +90,10 @@ public class BulkRegisterOdsInstancesModelValidator : AbstractValidator<BulkRegi
                                     database, apiModeProvider, databaseValidationService,
                                     databaseConnectionProvider, true);
 
-                                foreach (var record in model.DataRecords())
+                                var dataRecords = model.DataRecords();
+                                var newOdsInstanceToRegister = validator.GetNewOdsInstancesToRegister(dataRecords);
```
CSR2017, line 94: Please remove extra assignment to dataRecords  variable.

Application/EdFi.Ods.AdminApp.Management/OdsInstanceServices/OdsInstanceFirstTimeSetupService.cs
plioi: Regarding the idea of holding the filtered list on BulkRegisterOdsInstancesModel, I think that's a neat idea. Similar to the existing technique we use to ensure that the stream is only read one time. I'll elaborate, and you can correct me:

* BulkRegisterOdsInstancesModel could gain a new METHOD similar to DataRecords(). It would expose a private field that has a subset of _dataRecords in it. The new field and new method would start out null or empty.
* BulkRegisterOdsInstancesModelValidator already injects the DbContext, so it is free to use that to determine which ones are to be processed, and could "hold" that result by populating the new new field.
* Then the rest of the validator and the command can just look at the result of the new FilteredDataRecords method.

Is that the idea, @CSR2017? The model could even have a convenience method for the SkippedDataRecords(), and it would *calculate* that each time you call the method by just comparing _dataRecords against _filteredDataRecords. Something like: `return _dataRecords.Except(_filteredDataRecords)`.

CSR2017: @plioi 
Yes, thanks for adding more clarity.

leonardoSotoGap: @CSR2017 AND @plioi Thank you so much for your recommendations :) I learnt a lot about this PR and more from the code.
I already did the adjustments based on your comments, i expect everything is right now. Please let me know if you have questions or any concern there.

Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/OdsInstances/BulkRegisterOdsInstancesModel.cs
leonardoSotoGap: Done!!

Application/EdFi.Ods.AdminApp.Management/OdsInstanceServices/OdsInstanceFirstTimeSetupService.cs
```diff
@@ -65,6 +66,6 @@ private async Task AddOdsInstanceRegistration(OdsInstanceRegistration odsInstanc
         {
             _database.OdsInstanceRegistrations.Add(odsInstanceRegistration);
             await _database.SaveChangesAsync();
-        }
+        }
```
CSR2017, line 69: @leonardoSotoGap 
Please remove all the extra spaces.

CSR2017 (commented):
Please make sure to remove un-intended white space differences, which are causing the files getting added to PR. thanks

Application/EdFi.Ods.AdminApp.Management/Instances/RegisterOdsInstanceCommand.cs
```diff
@@ -3,6 +3,8 @@
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
 
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
```
CSR2017, line 8: @leonardoSotoGap 
Remove unused name spaces. Which is causing the changes on un-intended file.

Application/EdFi.Ods.AdminApp.Management/OdsInstanceServices/IOdsInstanceFirstTimeSetupService.cs
```diff
@@ -14,4 +15,4 @@ public interface IOdsInstanceFirstTimeSetupService
     {
         Task CompleteSetup(OdsInstanceRegistration odsInstance, CloudOdsClaimSet claimSet, ApiMode apiMode);
     }
-}
\ No newline at end of file
+}
```
CSR2017, line 18: @leonardoSotoGap 
No changes needed on this file.

Application/EdFi.Ods.AdminApp.Management/OdsInstanceServices/OdsInstanceFirstTimeSetupService.cs
```diff
@@ -1,8 +1,9 @@
-﻿// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: Apache-2.0
 // Licensed to the Ed-Fi Alliance under one or more agreements.
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using EdFi.Admin.DataAccess.Contexts;
 using EdFi.Ods.AdminApp.Management.Configuration.Claims;
```
CSR2017, line 9: Same as above

Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/OdsInstances/RegisterOdsInstanceModel.cs
```diff
@@ -3,6 +3,7 @@
 // The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 // See the LICENSE and NOTICES files in the project root for more information.
 
+using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using EdFi.Ods.AdminApp.Management.Database;
```
CSR2017, line 9: @leonardoSotoGap 
Please remove unwanted namespace.

Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/OdsInstances/BulkRegisterOdsInstancesModel.cs
```diff
@@ -53,16 +56,22 @@ private void EnsureReadStreamOnce()
     public class BulkRegisterOdsInstancesModelValidator : AbstractValidator<BulkRegisterOdsInstancesModel>
     {
         private readonly ILog _logger = LogManager.GetLogger("BulkRegisterOdsInstancesLog");
-       
+        private static AdminAppDbContext _database;
+        private static IDatabaseConnectionProvider _databaseConnectionProvider;
+        private static ApiMode _mode;        
+
         private bool UniquenessRuleFailed { get; set; }
```
CSR2017, line 63: @leonardoSotoGap 
I guess      private static AdminAppDbContext _database;
        private static IDatabaseConnectionProvider _databaseConnectionProvider;
        private static ApiMode _mode;   not used anywahere

CSR2017 (approved):

PR #88, opened 2020-12-01
[AA-1051] Fix issue when clicking cancel button in a add vendor modal.

leonardoSotoGap:
What was the problem?
Entering data in add vendor form and clicking cancel does not clear the form. If we opened the modal again the previous data and previous validations errors were still there.

Why was it happening?
The "@{ await Html.RenderPartialAsync("_AddVendorModal", new AddVendorModel()); }" instruction was listening the action from button with data-target ".add-vendor-modal". The problem is that this RenderPartialAsync instruction does not delete the previous modal instance, and for that reason the data was still there.

What was my solution?
After closing the modal, we reset all the fields and validations from the form, to avoid navigating to the server just to get a partial view with the new modal instance and affecting a few the performance of that process.


Application/EdFi.Ods.AdminApp.Web/Views/GlobalSettings/_AddVendorModal.cshtml
```diff
@@ -31,4 +31,14 @@ See the LICENSE and NOTICES files in the project root for more information.
...
         </div>
     </div>
-</div>
\ No newline at end of file
+</div>
+
+<script type="text/javascript">
+    $(function() {   
+        $(".add-vendor-modal").on("hidden.bs.modal", function () {            
+            $("#addVendorFormModal").get(0).reset()
```
saa14, line 39:
```suggestion
            $("#addVendorFormModal").get(0).reset();
```

saa14 (approved):

PR #99, opened 2020-12-10
[AA-1176] Display ODS / API version number in ODS instance pages of Admin App

leonardoSotoGap:
The idea of this Story was to add a little tag over the ODS instances pages with the current version of the ODS/API that the admin app is connected to.

Also it was added a label to the number at the right bottom of the footer to identify that number like Admin App version

I also did a little improvement getting the Admin App version. For that i added the cache approach there to avoid constantly go to server to get the same variable

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/Helpers/HtmlHelperExtensions.cs
+++ b/Application/EdFi.Ods.AdminApp.Web/Helpers/HtmlHelperExtensions.cs
@@ -23,7 +23,6 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
 using Property = EdFi.Ods.AdminApp.Web.Infrastructure.Property;
 using Preconditions = EdFi.Common.Preconditions;
-using System.Runtime.Caching;
 using EdFi.Ods.AdminApp.Management.Api;
 
 namespace EdFi.Ods.AdminApp.Web.Helpers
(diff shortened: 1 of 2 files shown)


Application/EdFi.Ods.AdminApp.Web/Helpers/HtmlHelperExtensions.cs
```diff
@@ -509,7 +511,7 @@ public static HtmlString ApplicationVersion(this IHtmlHelper helper)
             var assembly = Assembly.GetExecutingAssembly();
             var informationVersion = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
```
saa14, line 512: There is a good opportunity here to use caching as in OdsApiVersion and store the Admin App version as well.

Application/EdFi.Ods.AdminApp.Web/Helpers/HtmlHelperExtensions.cs
```diff
@@ -521,5 +523,17 @@ public static HtmlString ApplicationVersion(this IHtmlHelper helper)
                 label.Append(input).AppendHtml(icon);
             return label;
         }
+
+        public static HtmlString OdsApiVersion(this IHtmlHelper helper)
+        {
+            var odsApiVersion = MemoryCache.Default["odsApiVersion"];
```
saa14, line 529: I think the already established infrastructure for InMemoryCache can be used here
`var odsApiVersion = InMemoryCache.Instance.GetOrSet(
                "OdsApiVersion",
                () => new InferOdsApiVersion().Version(CloudOdsAdminAppSettings.Instance.ProductionApiUrl));
`

Application/EdFi.Ods.AdminApp.Web/Helpers/HtmlHelperExtensions.cs
leonardoSotoGap: @saa14 You are right thank you for your recommendation, i added cache approach there too. Thanks

Application/EdFi.Ods.AdminApp.Web/Helpers/HtmlHelperExtensions.cs
leonardoSotoGap: @saa14  yes, i agree with you is better to use what we already have in code for caching Thanks

saa14 (approved):

PR #107, opened 2020-12-16
[AA-1130] PART1: Admin app docker file system writes - resolve bulk upload hash cache folder

CSR2017:
@stephenfuqua 
1. Updated GetWorkingDirectry method on "LocalFileSystemFileUploadHandler.cs " to return custom directory path (default directory value specified on appsettings)
2. Updated app setting "BulkUploadHashCache" to have default value (C:\\ProgramData\\Ed-Fi-ODS-AdminApp\\BulkUploadHashCache)

Thanks


stephenfuqua (commented):
Code looks good. I will do some manual testing before merging it.

stephenfuqua (approved):
I found a bug in the way that the UI handles an error when the directory cannot be created. However, this is an existing bug that just happens to be exploited by this code. Therefore I'm satisfied with this pull request. Approving it - but waiting on merge until after the daily standup and conversation about this.

stephenfuqua (approved):

PR #109, opened 2020-12-18
[AA-1172] Adjust version number in database-manage powershell script

leonardoSotoGap:
As it was done in run-dbup-migrations powershell script, it was necessary to adjust the version in database-manager script.

The idea was to include a constant at the top of the file and reference it where it was needed instead of put string with different versions over all the script.


stephenfuqua (approved):
Looks good. I need to test it before merging.

leonardoSotoGap:
Thank you for your revision @stephenfuqua . Go ahead

PR #110, opened 2020-12-18
[AA 1195] Add "None" value to empty applications profile

leonardoSotoGap:
In Current version of the ODS/API  sample profiles were temporarily dropped from v5.1.0 - but coming back in v5.2.0.
For that reason it was defined to Add "None" value to profile field in application list view instead of showing up an empty or blank field.


stephenfuqua (approved):
Looks good. Will test before merging.

PR #118, opened 2020-12-31
[AA-1188] - IIS Rewrite module changes for AdminApp

saa14:
**Description**
- Added explicit web.config with iis rewrite url section.
- Removed UseHSTS and UseHttpsRedirection entries from StartUp.cs file.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/Properties/launchSettings.json
+++ b/Application/EdFi.Ods.AdminApp.Web/Properties/launchSettings.json
@@ -29,7 +29,6 @@
                 "ASPNETCORE_ENVIRONMENT": "SqlServer-YearSpecific"
             }
         },
-
         "District Specific (Postgres)": {
             "commandName": "IISExpress",
             "launchBrowser": true,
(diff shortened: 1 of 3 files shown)


stephenfuqua (changes requested):
I'm not entirely comfortable with the HSTS change yet, even if the Platform team already did the same. 

It might make sense when running as a Docker container with NGiNX in front of it, but probably not when running in IIS. I'm researching this in more detail.

If I recommend that we restore it then I'll file a ticket with the Platform team as well. Actively researching this today to try to avoid keeping this ticket blocked for very long.

Application/EdFi.Ods.AdminApp.Web/web.config
```diff
@@ -0,0 +1,31 @@
...
+            <action type="Redirect" url="https://{HTTP_HOST}/{R:1}" redirectType="Permanent" />
+          </rule>
+        </rules>
+        <outboundRules>
+          <rule name="Add Strict-Transport-Security when HTTPS" enabled="true">
+            <match serverVariable="RESPONSE_Strict_Transport_Security" pattern=".*" />
+            <conditions>
+              <add input="{HTTPS}" pattern="on" ignoreCase="true" />
+            </conditions>
+            <action type="Rewrite" value="max-age=31536000" />
```
stephenfuqua, line 25: OK, this is the IIS-level setting for HSTS. Now I'm satisfied that we still have it in place. Just need to insure, next, that we're configuring NGiNX also to use HSTS.

stephenfuqua (approved):
After further review everything looks good to me. Have not tested this therefore not merging it yet.

stephenfuqua:
@CSR2017 @saa14 we should test this security header in IIS as part of the review/qa process, see https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.html

stephenfuqua (changes requested):
The presence of the web.config file appears to be making it impossible to run one of the launch profiles. For example, every time I try to run "Shared Instance (SQL Server)" it continues trying to run the Azure (default) settings. I can see this in the error messages in the Visual Studio output pane and in the fact that the error logs are being written to the Azure location. Furthermore, the application just has a big warning message that an error occurred. When I delete the web.config then everything is fine. I don't know why this would happen, but it seems that some kind of rework will be necessary.

Application/EdFi.Ods.AdminApp.Web/Properties/launchSettings.json
```diff
@@ -58,15 +57,53 @@
                 "ASPNETCORE_ENVIRONMENT": "Docker-SharedInstance"
             }
         },
-
-
-        "EdFi.Ods.AdminApp.Web": {
+        "mssql-district": {
```
stephenfuqua, line 60: These new launch profiles allow you to run the application from the command line or in Visual Studio, using Kestrel directly instead of running it under IIS Express.

```bash
dotnet run -c debug --launch-profile mssql-shared
```

stephenfuqua (approved):

PR #132, opened 2021-01-15
[AA-1228] Improve user experience when ODS version lookup throws exceptions

plioi:
This prevents ODS version lookup failures from breaking the surrounding page: the version is shown when available, and exceptions are logged, but the page is given the opportunity to otherwise load.


saa14 (approved):

saa14:
**Dev Testing**
- Tested locally to see the code in action
- Checked the local Admin App logs for the logged error.

PR #133, opened 2021-01-15
[AA-1006] - Take nuget feed as argument for Admin App Installer

saa14:
**Description**
Added the PackageSource/Nuget feed parameter to be passed into the Get-NugetPackage function in Install-EdFiOdsAdminApp.psm1 and prep-installer-package.ps1


EdFi.Suite3.Installer.AdminApp/prep-installer-package.ps1
```diff
@@ -32,6 +32,7 @@ $parameters = @{
     PackageName = "EdFi.Installer.AppCommon"
     PackageVersion = $dependencyVersions.AppCommon
     ToolsPath = "C:/temp/tools"
+    PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json"
```
stephenfuqua, line 35: When I wrote the ticket, I was thinking that this script is parameterized. Just hard-coding it here doesn't help very much - if I want to change the feed URL then I need to manually modify the script. We should probably parameterize the version number from line 17, but we can leave that for another day.

So, that said, can you add a parameter block for the PackageSource at the top of the file, and move the default value up to the parameter? Thus if I need to change from TeamCity, for example, I could do that with a setting instead of changing the script itself.

stephenfuqua: Argh, I was distracted and didn't scroll to the top of the file. We have a parameters block already. Just need to add this parameter there please, next to packages directory. If it is not too much trouble, this release will be much better if we can go ahead and move the package version to the parameter block as well, continuing to have the current default value.

saa14: 👍 I was thinking of doing the same.

stephenfuqua (changes requested):
Small change request

stephenfuqua (approved):

PR #136, opened 2021-01-26
[AA-1239] - Display external API URLs under Docker

saa14:
**Description**

**NOTE: This PR is a companion of the  [Docker counterpart](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Docker/pull/50)**

- Added ApiExternalUrl appsetting to set the ods/api url. This appsetting is only supposed to be used for displaying the correct url in docker environments.
- Added a private method GetApiUrlForDisplay to replace the production Api Url with the specified ApiExternalUrl if present. Refactored the controller actions to use the method.
- Updated the build script and BuildDockerDevelopment script to take the new appsetting of ApiExternalUrl into account.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/Controllers/ApplicationController.cs
+++ b/Application/EdFi.Ods.AdminApp.Web/Controllers/ApplicationController.cs
@@ -74,14 +74,12 @@ namespace EdFi.Ods.AdminApp.Web.Controllers
             var apiUrl = (await _apiConnectionInformationProvider.GetConnectionInformationForEnvironment())
                 .ApiServerUrl;
 
-            apiUrl = GetApiUrlForDisplay(apiUrl);
-
             var model = new ApplicationsIndexModel
             {
                 OdsInstanceSettingsTabEnumerations =
                     _tabDisplayService.GetOdsInstanceSettingsTabDisplay(OdsInstanceSettingsTabEnumeration.Applications),
                 OdsInstance = _instanceContext,
-                ProductionApiUrl = apiUrl
+                DisplayApiUrl = GetApiUrlForDisplay(apiUrl)
             };
 
             return View("Index", model);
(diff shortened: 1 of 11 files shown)


BuildDockerDevelopment.ps1
```diff
@@ -18,6 +18,7 @@
 
 $p = @{
         ProductionApiUrl = "http://api"
+        ApiExternalUrl = "https://localhost:5001"
```
plioi, line 21: Please include similar example over in the documentation comment of build.ps1.

Application/EdFi.Ods.AdminApp.Web/Controllers/ApplicationController.cs
```diff
@@ -214,19 +224,32 @@ public ActionResult RegenerateSecret(RegenerateSecretModel model)
...
+                ApiUrl = apiUrl
             });
         }
 
         private List<string> GetClaimSetNames()
         {
             return _getClaimSetNamesQuery.Execute().Except(CloudOdsAdminApp.SystemReservedClaimSets).ToList();
         }
+
+        private static string GetApiUrlForDisplay(string apiUrl)
```
plioi, line 247: I think this should be a get-only property on CloudOdsAdminAppSettings.Instance. A clue that it belongs there is that the body of this method asks that object for a few values and then calculates the result with no need for other inputs.

plioi: I also prefer not to need mutation of the input parameter here.

stephenfuqua:
@saa14 please bump the version to 2.1.0 in `.teamcity\_Self\Project.kt`.

plioi:
@stephenfuqua You mean to 2.1.1, correct? The current value is 2.1.0 set a few months back.

plioi (approved):

PR #139, opened 2021-02-05
[AA-1180] remove deprecated reset to minimal template

plioi:
The "Setup" tab, hidden in on-prem deployments, was intended to be removed back when we removed the Staging vs Production concepts. Instead of being removed, it was left in a state where it was accessible in Cloud ODS deployments, though it did no meaningful work and had little to say to the user. It's no longer necessary since Descriptors would already be present at database deployment time. So, we had a fairly large set of code that was in practice unreachable and would not perform meaningful work if it ever were reached.

This PR phases out the deprecated code that was intended to be removed previously. Although you may want to review the overall diff as a whole, it may help to see the technique used by scanning the individual commits in order. In general, I removed code "bottom up" starting with things we knew had no meaningful use, relying on R# and compiler errors to guide me towards removing *all* the unreachable code.


saa14 (approved):

saa14:
**Developer testing**
- Went through the individual commits
- Local build testing
- Unit tests pass successfully

PR #140, opened 2021-02-09
[AA-1168] remove deprecated updates tab

plioi:
This removes the deprecated Updates tab, which was only visible for Cloud ODS deploys, performed no changes, and displayed information which was in fact inaccurate. Rather than maintain it further, it is meant to be removed.

This involved touches to the Home screen button display logic, which was found to be needlessly complex (some decisions of item inclusion was handled in the intended service class, while some was being handled in the view). This simplifies the logic to one place so that it is easy to map code to item visibility requirements.


saa14 (approved):

saa14:
**Developer testing**
- Went through the individual commits
- Local build testing (including different configurations for Home Screen testing)
- Unit tests pass successfully

PR #147, opened 2021-02-25
[AA-932] A user must be both logged in and have a matching Users record to access protected pages.

plioi:
Using MVC Core Authorization Policies, ensure that users are authorized if they are both authenticated (a valid login cookie exists) *and* a corresponding User record still exists for that user.

This gives a consistent experience for two interesting cases:

* A developer logs in, and while the cookie is still considered valid the developer resets the contents of their Users table. The cookie is still considered valid for Authentication but here we apply more strict Authorization and send them to the Login page again.

* A user logs in, and while they are interacting with the site a Super Admin deletes their account. As the first user tries to continue interacting, they are rightly sent back to the Login page.

Useful links which led to this solution:

* https://andrewlock.net/setting-global-authorization-policies-using-the-defaultpolicy-and-the-fallbackpolicy-in-aspnet-core-3/
* https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-5.0

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/Startup.cs
+++ b/Application/EdFi.Ods.AdminApp.Web/Startup.cs
@@ -23,6 +23,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using FluentValidation.AspNetCore;
 using Hangfire;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Identity;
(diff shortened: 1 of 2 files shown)


plioi:
@saa14 You found that the UI presented as if the user was still logged in, when they are kicked to the login page for having no matching record. I've corrected for this by signing them out upon the user record lookup failure. I use the same technique as what happens when a user explicitly logs out.

saa14 (approved):

PR #148, opened 2021-03-02
AA-1182 generalize connection string token replacement to match ODS token replacement

plioi:
This generalizes connection string `"{0}"` replacement patterns so that, just like the ODS implementation itself, we too allow for token replacement for *all* supported API modes, not just the multi-instance modes.

Specifically, just like the ODS, we now support a connection string for Shared Instance mode with the template database name `"EdFi_{0}"` which naturally gets substituted to `"EdFi_Ods"`. The primary effect is that our various sample appsettings files become that much more alike. The ability to use some alternate specific database name is unharmed, which for example happens in Cloud ODS deployments.

I began with characterization test coverage of the affected ConnectionStringService and then extended the tests as the handling of Shared Instance evolved.

This also includes an update to developer-facing documentation to better match the typical moves a developer would make as they switch between modes. Superficial edits here also bring the 3 similar sections into alignment so they are less likely to drift again and so they use similar phrasing for similar things.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Management.Tests/Services/ConnectionStringServiceTests.cs
+++ b/Application/EdFi.Ods.AdminApp.Management.Tests/Services/ConnectionStringServiceTests.cs
@@ -32,6 +32,23 @@ namespace EdFi.Ods.AdminApp.Management.Tests.Services
         }
 
         [Test]
+        public void Sandbox()
+        {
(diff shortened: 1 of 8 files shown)


saa14 (approved):

saa14:
**Developer testing**
- Tested the ODS API connection under SharedInstance, YearSpecific and DistrictSpecific modes
- All the unit tests pass successfully

PR #149, opened 2021-03-08
[AA-960] - Use efficient paging controls when handling results from the ODS/API

saa14:
**Description**
- Added Page.cs and corresponding tests to handle server side paging functionality. This is similar to the DataImport paging functionality.
- Added HtmlHelperExtensions methods to create the html tags for the paging control.
- Added an overload for GetAll method to specify the starting offset and limit (number of items to get) to facilitate the paging flow by getting only the items to be displayed on the page. An extra item is added to indicate that more items are present to fetch.

**Education Organizations Tab Paging**

- Added GetLocalEducationAgenciesByPage method to facilitate paging on EducationOrganizations view.
- Refactored the EducationOrganizationViewModel to use the PagedList class for LocalEducationAgencies and added the LocalEducationAgenciesPageNumber to keep track of the page number.
- Refactored Page.cs to use Task<IEnumerable<T>> as the return type for the getApiRecords method.Added PagingControl to _EducationOrganizations.chtml and refactored Model.LocalEducationAgencies to use Items property of the PagedList.
- Added PagingControl to _EducationOrganizations.cshtml and refactored Model.LocalEducationAgencies to use Items property of the PagedList.Added paging control JS functions to attach paging behavior to div tags with ajax-content css class. Wrapped the first call to EducationOrganizationList in a div with the ajax-content class to allow the paging behavior.
- Refactored EducationOrganizationControllerTests to use the GetLocalEducationAgenciesByPage method instead of GetAllLocalEducationAgencies.Refactored the Page.cs unit tests to use the async version of getApiRecords instead of the synchronous version

**Applications Tab Paging**

- Added VendorApplicationsIndexModel to add paging to the vendors on Applications tab. Refactored _Applications.cshtml to use VendorApplicationsIndexModel.
- Added PagingControl to _Applications view and added the ajax-content div to Index to allow for div replacement on paging.
- Refactored ApplicationList() to add a local function and pass that function to Fetch.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
+++ b/Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
@@ -475,6 +475,7 @@ var replacePagedContent = function (event) {
 
     $.ajax(url).done(function (data) {
         $btn.closest('.ajax-content').html(data);
+        GlobalInitialize();
     });
 };
 
--- a/Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
+++ b/Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
@@ -480,6 +480,9 @@ var replacePagedContent = function (event) {
 };
 
 function AttachDefaultPagingBehavior() {
-    $(document).on('click', '.navigate-previous-page', replacePagedContent);
-    $(document).on('click', '.navigate-next-page', replacePagedContent);
+    $('.navigate-previous-page').off("click");
+    $('.navigate-previous-page').on("click", replacePagedContent);
+
+    $('.navigate-next-page').off("click");
+    $('.navigate-next-page').on("click", replacePagedContent);
 };   
--- a/Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
+++ b/Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
@@ -46,6 +46,7 @@ function GlobalInitialize() {
     SetupPanelToggle();
     LoadAsyncActions();
     ClaimSetWarningMessage();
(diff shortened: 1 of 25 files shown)


plioi:
As discussed 1-to-1:

- [ ] Use `IReadOnlyList<T>` rather than `IEnumerable<T>` for the page's Items collection to communicate that it is already "realized" rather than having "lazy" semantics.
- [ ] Apply async/await keywords instead of accessing Task.Result.
- [ ] Explore whether the paging controls helper can accept a single Url.Action(method, controller) and the current page number, assembling the two similar URLs for you.

plioi:
@saa14 I rebased and applied a few updates. Before getting into the following feedback, be sure to delete your local branch, pull while on 'main', and then check out so that you arrive at e391bf498aecad9.

Then:

1. Scan the commits I added so we're on the same page.
2. The 2 affected view models gained both a page number property and the `PagedList<T>` object. But, PagedList contains the PageNumber too, so I think we only need the PagedList on these view models. Looks like this also means that the PagingControl helper could accept one fewer arguments, since the final argument already contains the page number. In addition to being less code, this also means that we're fully benefitting from the fact that `Page<T>` fixes up bad page 0 values to 1.
3. I don't have feedback yet on the ApplicationController, but I'd like to hear in your own words which aspects of this will naturally still be slow in the *worst* case due to fetching all items before we hit Skip and Take.

Application/EdFi.Ods.AdminApp.Web/wwwroot/Scripts/site.js
```diff
@@ -466,6 +467,27 @@ function UpdateWarningTimer() {
 
 };
 
+<<<<<<< HEAD
```
plioi, line 470: Looks like a merge conflict resolution was not saved as expected. In addition to removing this line, double check that the remaining code change here looks like you originally intended.

plioi (approved):

PR #153, opened 2021-03-11
[AA-162] Configure School Year

plioi:
This allows the user to set the ODS instance-level School Year setting, managed in the ODS edfi.SchoolYearType table. The table acts as a lookup for available years as well as having one record marked as the current year. This feature lets the Admin App user witness the current selection and to change that selection.

In Shared Instance mode, the year selection is visible and editable in the header of the Instance Settings screen. In District/Year Specific modes, it is visible and editable from the rows of the instance selection screen.

There are special considerations for Year Specific mode. At instance registration time, we default the instance's year selection to the single year that the instance is dedicated for, saving the user from (likely) ever needing to edit it themselves. However, we do still offer the controls, primarily so a user could *correct* the system if the SchoolYearType table were altered outside of the app. When a user goes to edit the year in Year Specific mode, not only do we put up a warning but we suggest exactly what value is *expected* for that instance. This warning and suggestion also helps to guide the user away from an "off by one" mistake as "Current Year 2021" really corresponds with a "2020-2021" school year.


saa14 (approved):

saa14:
**Dev Testing**
- All tests pass successfully
- Tested the functionality across sharedinstance, yearspecific and districtspecific modes

PR #154, opened 2021-03-12
[AA-1148] - Disable the rest of the table while editing a row on the resource editor

saa14:
**Description**
- Added css classes for selection to the table cells with edit and delete icons on the ResourceEditor.
- Added styles for disabled table cells. Added style to make the edit-resource <a> tag similar to other icons.
- Added AddTooltip and Removetooltips functions to site.js for adding/removing tooltips from a tag.
- Added disableForEdit and enableAfterEdit with helper functions to enable and disable the table cells depending upon the row state while editing on the ResourceEditor.


plioi:
We discussed leaving the delete buttons enabled, as there didn't seem to be a downside to it. However, the user can get "stuck" if they perform the following sequence of steps:

1. From the main list of claim sets, select one to go to the edit screen.
2. Within one row of the edit screen, click the edit icon to make that row editable. 
3. Within the *same* row, click the delete button.
4. In the deletion modal, confirm the deletion.
5. The remainder of the controls are now locked, since the single active row is now gone and cannot be saved. The user would have to leave the editor completely and come back in order to do more work.

Rather than doing anything more complicated, let's just disable all the delete buttons for the same reason that we disable all of the "(i)" auth strategy buttons.

plioi:
Add a new row, save, then back to edit mode. The Delete tooltip is not set.

I knew to check for this after seeing that we seem to have a lack of symmetry in the changes. I see the cshtml changes apply .edit-resource-button, .edit-auth-strategy, and .delete-resource-claim to all the nodes when rendered, but in the JS we see variations on when these css classes are mentioned/introduced on the javascript side.

For instance, when the javascript introduces .edit-resource-button around line 23, why not the others? For instance, when the javascript introduces .edit-auth-strategy near line 225, why not the others? These two examples might be right, but the different treatments of the 3 and the lack of js-introduced .delete-resource-claim led me to find this bug.

plioi (approved):

PR #155, opened 2021-03-17
AA-991 Refactor TeamCity config to create two configs for AdminApp.Web and AdminApp.Installer

andonyns:


.teamcity/Installers/buildTypes/BuildAdminAppInstaller.kt
```diff
@@ -47,4 +50,8 @@ object BuildAdminAppInstaller : BuildType ({
             }
         }
     }
+
+    requirements {
+        equals("teamcity.agent.name", "INTEDFIBUILD3")
```
stephenfuqua, line 55: The only reason I can think of to restrict to this build agent is for using Advanced Installer. But you don't need Advanced Installer any more. Can this restriction be removed?

.teamcity/_Self/vcsRoots/EdFiOdsImplementation.kt
```diff
@@ -0,0 +1,25 @@
...
+    branch = "%git.branch.default%"
+    branchSpec = "%git.branch.specification%"
+    userNameStyle = GitVcsRoot.UserNameStyle.FULL
+    checkoutSubmodules = GitVcsRoot.CheckoutSubmodules.IGNORE
+    serverSideAutoCRLF = true
+    useMirrors = false
+    authMethod = password {
+        userName = "%github.username%"
+        password = "%github.accessToken%"
+    }
```
stephenfuqua, line 24: You can change this to anonymous now.

.teamcity/Installers/buildTypes/BuildAdminAppInstaller.kt
andonyns: The reason I did this was since no agent was picking up the build and it said there was no compatible agents. I'll go back to the fork and test this in the build

.teamcity/_Self/vcsRoots/EdFiOdsImplementation.kt
andonyns: Thanks. I'll changing it

.teamcity/Installers/buildTypes/BuildAdminAppInstaller.kt
stephenfuqua: You've probably already [noticed this](https://intedfitools1.msdf.org/admin/editRequirements.html?id=buildType:Experimental_OdsTools_EdFiOdsAdminApp_Deploy), but just in case... for the Deploy task you need an `octopus.server` variable. You should define that in the `ODS Tools` project - the parent of the project with versioned settings.

PR #160, opened 2021-03-22
[AA-991] - Move EdFi.Installer.AdminApp Build to New Area

andonyns:
Creating two build configuration for Admin App, one for the Web package and another one for the Installer package.

The Deploy process will now be part of the Installer process rather than the Web process, this was moved to a new area but the change required will be addressed in an upcoming ticket


.teamcity/Installers/Project.kt
```diff
@@ -0,0 +1,23 @@
...
+
+object AdminAppInstallerProject : Project({
+    name = "Admin App Installer"
+    id = RelativeId("AdminAppInstaller")
+    description = "ODS Admin App Installer Build Configurations"
+
+    params {
+        param("project.name", "EdFi.Suite3.Installer.AdminApp")
+        param("adminAppInstaller.version", "2.2.0")
+        param("adminAppWeb.version", "2.2.0-pre0008")
```
stephenfuqua, line 18: Should the build number be driven by the TeamCity build counter, rather than hard-coded?

I see `2.2.0` in two places in this file, and in another file lower down. Would it make sense to consolidate those into a single project-level parameter in `.teamcity/_Self/Project.kt`?

.teamcity/Installers/buildTypes/Deploy.kt
```diff
@@ -1,88 +1,88 @@
...
+
+object Deploy : BuildType ({
+    name = "Deploy"
+    description = "Creates a release in Octopus Deploy and triggers its deployment to the test server"
+    
+    type = BuildTypeSettings.Type.DEPLOYMENT
+
+    params {
+        param("octopus.environment", "Integration")
+        param("octopus.channel", "SharedInstance v5.1.x")
```
stephenfuqua, line 21: Is this truly specific to version 5.1.x?

.teamcity/Installers/Project.kt
andonyns: Thanks. I'll update these parameters

.teamcity/Installers/buildTypes/Deploy.kt
andonyns: We are following a structure for the Octopus Deploy of ODSMode_version_AdminApp_version to have multiple AdminApp versions deployed into the server, that's why the deploy is for the shared instance channel only. 

This change was to be able to create a successful Deploy to Octopus in order to have the process working but we have [AA-1008](https://tracker.ed-fi.org/browse/AA-1008) in the sprint in order to improve this process by creating 3 deploys, one per mode and obtain the AdminApp.Web version from created package.

.teamcity/Installers/Project.kt
andonyns: I updated the parameters to get the value from the Project.kt file

CSR2017 (approved):

PR #163, opened 2021-03-30
[AA-1276] Refactor - encryption and decryption technic on Admin App

CSR2017:
@plioi @saa14 
1. Updated code to have just AES encryption service.
2. Removed Optional entropy and Encryption protocol from appsettings and build script 
3. Added db migration script for adding IsEncrypted column to SecretConfiguration and AzureSqlConfiguration tables
4. Update Odssecretconfigurationprovider to handle IsEncrypted column while reading/ writing the configuration value
5. Update Admin app installer to auto generate Encryption key value during the installation

Thanks


plioi:
1. Be sure to get latest and scan my updates to the tests. There are two changes here. First, the *preexisting* test setup code was needlessly complex with mocking, and I simplified that now that we can just have our test-specific encryption key in the test-specific appsettings file. This way, we're really exercising more of the real IoC and settings lookup code. Second, I added assertions around the new database column being true or false at each step.
2. The Postgres migration script will fail. If you compare the two database engine's folders of migrations, you'll see that the Azure specific table is never created in Postgres since it would never be needed at runtime.
3. The root appsettings.json file correctly gained the new appsetting, but understandably it leaves it with a placeholder value. I agree, since that will force us to do a good job of populating it differently in each deployment path like you do in the install scripts. But, this means that if I clone the repo and try to run, I'll get surprising exceptions when it tries to use that placeholder as a usable key. I really like leaving it like this for the root appsettings file, but can we place a developer-specific value in the different developer-facing appsettings.*.json files? Then development can always work without having to know about it.

plioi:
On that last point, it might be enough to place it in the launchSettings.json file, which is just for development, in the environmentVariables sections. I think that a correctly-named environment variable will get picked up similar to our json files. Then it would be even more clear that this is just a developer-facing value and not an actual secret.

saa14 (approved):

PR #168, opened 2021-04-13
[AA-1293] Anti-forgery token error on docker environment

CSR2017:
@plioi @saa14 
1. Implemented custom IXmlRepository to override the default key management on data protection.
2. Added DataprotectionKeys table
https://tracker.ed-fi.org/browse/AA-1293
Thanks


Application/EdFi.Ods.AdminApp.Web/Startup.cs
```diff
@@ -76,6 +79,15 @@ public void ConfigureServices(IServiceCollection services)
                                     .GetCustomAttribute<System.ComponentModel.DataAnnotations.DisplayAttribute>()?.GetName();
                         });
 
+            services.AddSingleton<IXmlRepository, DataProtectionRepository>();
```
plioi, line 82: If this is a singleton, then the DbContext inside it will accidentally live forever, too. Can this be transient or scoped instead, and still work?

plioi, line 82: Is there a reason we couldn't go with the EF Core option here instead of the "xml" approach? https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/implementation/key-storage-providers?view=aspnetcore-5.0&tabs=visual-studio#entity-framework-core

CSR2017: @plioi 
Updated the code to use custom dataprotection database context. Which reduces the overhead of maintaining the repository, model, and handling the scope.

plioi:
Wow, that change turned out very nice. I'm going to reassign review to @saa14 with the following guidance to him:

- [ ] Start by merging from `main` into the PR branch to be sure you're up to date.
- [ ] Experiment with removing the new package reference from the Web project. Since references are "transitive" in modern csproj files, we appear to already get this "for free" since Web references Management which references the new library.
- [ ] Normal line-by-line code review. I took a quick look but I recommend contrasting the changes with the guidance at https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/implementation/key-storage-providers?view=aspnetcore-5.0&tabs=visual-studio#entity-framework-core - Clearly there were a few places where this code needed to diverge from that overly-simple example, to match *our* patterns for things like Postgres vs SQL Server. See also Suganya's original analysis attached to the ticket. It at looks good at a quick glance but please take a close look mostly to spread around knowledge of the issue across the team.

saa14 (approved):

PR #170, opened 2021-04-20
[AA-1288] - Add the Production Google Analytics property

saa14:
**Description**

Added the production GA property to the root appsettings.json GoogleAnalyticsMeasurementId appsetting.


plioi (approved):

plioi:
As discussed elsewhere, these values are not in fact *secrets* so we're including it here in the root appsettings so that it arrives for real world users upon installation. Previous PR establishes the dev/test value for local runs, and a separate ticket will ensure this also arrives in Docker runs.

@saa14 I've approved of the PR. Merge at your convenience. Be sure to coordinate with @andonyns to be sure we're spending most of our time hitting the Dev/Test value in his deployment environments, but that a realistic end user installation arrives with the intended production value.

PR #171, opened 2021-04-22
[AA-1288] - Move Production GA property Measurement ID to TeamCity

saa14:
**Description**
- Removed the production GA property measurement ID from appsettings.json as the value is now shifted to TeamCity.
- Added PopulateGoogleAnalyticsAppSettings command to the build script to update appsettings.json with the GoogleAnalyticsMeasurementId value provided. Refactored build script to take in a GoogleAnalyticsMeasurementId parameter for this command.
- Added a build step to the BuildAndTest template to populate the GA measurement ID before proceeding to create the nuget package.

### DIFF
--- a/.teamcity/Web/templates/BuildAndTestTemplate.kt
+++ b/.teamcity/Web/templates/BuildAndTestTemplate.kt
@@ -32,6 +32,16 @@ object BuildAndTestTemplate : Template({
                 """.trimIndent()
             }
         }
+        powerShell {
+            name = "Populate Google Analytics AppSettings"
+            formatStderrAsError = true
+            executionMode = BuildStep.ExecutionMode.RUN_ON_SUCCESS
+            scriptMode = script {
+                content = """
+                    .\build.ps1 -Command PopulateGoogleAnalyticsAppSettings -GoogleAnalyticsMeasurementId "%adminApp.googleAnalyticsMeasurementId%"
+                """.trimIndent()
+            }
+        }
         // Note: there are no artifact rules on this template, so that the NuGet packages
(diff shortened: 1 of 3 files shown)


plioi:
@saa14 first, I did do some rebasing and adjustments with this branch, so before pulling or doing more work, I recommend deleting your local branch and then checking this one out from scratch.

Next, I expect this to throw exceptions if the Product Improvement feature is enabled and the user runs the TransformLoad console application. In that case, `HttpContext.Current` will be null and `Telemetry` will throw. You can quickly experiment with that yourself: change the TransformLoad Main() method to do only the following, and then run it under the debugger:

```cs
public static void Main()
{
            var userName = HttpContext.Current.User.Identity.Name;
            return;
}
```

It would be tempting to just introduce a ? in the property accessing chain when we mention HttpContext in Telemetry (ie. `HttpContext.Current?.Request.Url.Host`), but note that although that's a good start it would in turn lead to exceptions when the `userName` property is parsed, so be sure to take care with the `userName` being possibly-null as well.

What's nice is that since this is able to assume the .NET Framework for now, the `HttpContext` class *is* ok to mention, it's just that `HttpContext.Current` will be null in the console application. Since we're talking about a background process here, it's ok if these usages of Telemetry result in "(not set)" being reported to Google Analytics. So this really is just about suppressing null reference exceptions.

I strongly recommend:

* Reproduce the error. It's easy enough to set up a run of TransformLoad against the publicly hosted demo ODS instances, which might simplify setup.
* Then update Telemetry to be more pessimistic about HttpContext.Current being possibly-null.
* Try the repro scenario again, witnessing no more exceptions.

plioi (approved):

PR #191, opened 2021-07-22
[AA-1286] Removing 3.3.0-a Schema folder

andonyns:


plioi (approved):

plioi:
After discussing with Andony, we determined this folder cannot be safely removed. There is no harm in keeping it, though.

The publicly released ODS / API is version 5.2.14406, released via Azure Artifacts on March 24 2021. The ODS / API fix which suggested we could remove this folder from AA was merged here ( https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS/pull/308 ) on April 12 2021. So, Andony's testing of this AA branch against the publicly released ODS shows the original symptom remains.

The *hope* had been that *if* the ODS side fix were included for all end users of 5.2.*, then and only then could AA remove this folder as no longer necessary. Since the ODS fix will be in place 5.3 forward, we need to keep this folder in AA to ensure all users succeed. The fix in the ODS does allow AA to stop adding *new* data standard folders moving forward, but we will keep this one in place to ensure backward compatibility with ODS 5.2.

PR #199, opened 2021-08-16
[AA-1235] - Fixed error message when importing non JSON file while importing ClaimSets

saa14:
**Description**
Added file extension check for the ImportFile property in ClaimSetFileImportModel. Added corresponding unit test for the new check.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/ClaimSets/ClaimSetFileImportModel.cs
+++ b/Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/ClaimSets/ClaimSetFileImportModel.cs
@@ -44,9 +44,9 @@ namespace EdFi.Ods.AdminApp.Web.Models.ViewModels.ClaimSets
                         {
                             var validator = new SharingModelValidator(securityContext, context.PropertyName);
 
-                            if (Path.GetExtension(model.ImportFile.FileName)?.Substring(1) != "json")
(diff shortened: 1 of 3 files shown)


Application/EdFi.Ods.AdminApp.Web/Models/ViewModels/ClaimSets/ClaimSetFileImportModel.cs
```diff
@@ -42,7 +43,15 @@ public ClaimSetFileImportModelValidator(ISecurityContext securityContext)
                         .SafeCustom((model, context) =>
                         {
                             var validator = new SharingModelValidator(securityContext, context.PropertyName);
-                            context.AddFailures(validator.Validate(model.AsSharingModel()));
+
+                            if (Path.GetExtension(model.ImportFile.FileName)?.Substring(1) != "json")
```
plioi, line 47: There are two problems with this condition.

First, if the file has no extension, GetExtension() returns the empty string and then the Substring attempt throws an exception.

Second, this condition is too strict for valid JSON files. "example.JSON" with capital letters is rejected

Let's change it to:

`if (Path.GetExtension(model.ImportFile.FileName)?.ToLower() != ".json")`

Note we're calling ToLower *and* including the period in the comparison so we can exclude the Substring call. I think I'd also prefer the error message to say "*.json" rather than a bare "json". This way it's clear we're indicating the file extension and not directly talking about the format of the file content.

plioi (approved):

PR #204, opened 2021-08-28
[AA-1354] Claim set registration updates

jamessmckay:


plioi:
In case you're not able to reach the TeamCity builds, 4 integration tests are failing here. See https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/blob/main/docs/developer.md#build-script to repro these locally.

1. BulkRegisterOdsInstanceCommandTests.BulkShouldNotRegisterOneOdsInstancePreviouslyRegistered
2. BulkRegisterOdsInstanceCommandTests.ShouldRegisterOneNewOdsInstanceInBulk
3. RegisterOdsInstanceCommandTests.ShouldRegisterDistrictSpecificOdsInstance
4. RegisterOdsInstanceCommandTests.ShouldRegisterYearSpecificOdsInstance

OdsInstanceFirstTimeSetupService.CreateAndSaveApiKeyAndSecret (somewhere near line 59) throws null reference exceptions in at least three of these, though the TeamCity captured logs are a little garbled by Shouldly running in Release mode. My guess would be something caused CreateAdminAppInAdminDatabase to return null, but I have not tried to catch it under the debugger.

What I see in the actual code change looks like a merge resolution mixup. See how it's leaving the original `newInstance` variable in place though it seems that was meant to be replaced in favor of the new variable `odsInstance` variable, but then still returns `newInstance.Id` at the end, which is probably Guid.Empty now.

plioi:
I'm also surprised by the remaining change. @CSR2017 would it be valid for the OdsInstanceRegistration.Name to be only the numeric suffix? If so, how has anything worked to date?

stephenfuqua:
@pmcvtm can you evaluate if this is still relevant, and if not, close it? Thanks.

pmcvtm:
This one looks safe to close. Besides moving to `aysnc`, we have introduced a separate property (`OdsInstanceRegistrations.DatabaseName`) derived from the `NumericSuffix`, which appears to be part of the intent here.

@jamessmckay if I am missing something or you're still facing an issue please open a new PR from latest which addresses it! As an added bonus, we will see test failures right here in GitHub.

PR #205, opened 2021-08-28
[AA-1338] Hangfire brings down Postgres if not managed correctly

jamessmckay:


plioi:
I don't know of the context, but @jamessmckay and @CSR2017 but something to look out for here: we already have class HangFireInstance which seems to be making a few similar statements about what database engine we're connected to and how to establish sotrage for Postgres and SQL Server, so if this addition to Startup is right I suspect the old code is at best dead and at worst conflicting/redundant. Suganya, what do you think about the timeout durations with respect to how long these jobs tend to run in production?

stephenfuqua:
@pmcvtm can you evaluate if this pull request is still relevant? And if not, can you go ahead and close it please?

pmcvtm:
Sounds like _some_ of this may have already been addressed, or was an edge case. I am closing this but @jamessmckay if you are still having an issue and have a solution or proposal that meets your need, let us know.

PR #206, opened 2021-08-28
[AA-1336] Expand instance name field to 1024 characters

jamessmckay:


plioi:
The 3 failing integration tests are:

1. AddApplicationCommandTests.ShouldNotAddIfNameNotWithinApplicationNameMaxLength
2. EditApplicationCommandTests.ShouldNotEditIfNameNotWithinApplicationNameMaxLength
3. RegisterOdsInstanceCommandTests.ShouldNotRegisterInstanceIfOdsInstanceNameNotWithinApplicationNameMaxLength

Though I'm more concerned about the change than just making the tests switch to green.

1. Looking at the original code, the "51" in AddApplicationCommandTests was meant to be 1 *more* than the available room.
2. Was this meant to correspond with a schema change? Is the schema different across versions? We have a backward compatibility rule back to ODS 3.4.
3. Tracing usages of the ApplicationName we're validating here, it seems to correspond with the following (on the 5.2 side of our backward compatibility range, at least):

    Users.FullName (nvarchar(max))
    ApiClients.Name (nvarchar(50))
    Applications.ApplicationName (nvarchar(max))

So the full story will need to deal with that ApiClients size limit at least.

pmcvtm:
This change was addressed in depth with PRs #227 and #233. I am closing this PR. Please open a new one if you find differently.

PR #208, opened 2021-09-01
[AA-1331] Upgrade-Aware Installation

CSR2017:
@stephenfuqua @plioi @saa14 
Please help review the changes.

### DIFF
--- a/EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
+++ b/EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
@@ -1,4 +1,4 @@
-# SPDX-License-Identifier: Apache-2.0
+﻿# SPDX-License-Identifier: Apache-2.0
 # Licensed to the Ed-Fi Alliance under one or more agreements.
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
@@ -324,6 +324,149 @@ function Install-EdFiOdsAdminApp {
     }
 }
 
+function Upgrade-EdFiOdsAdminApp {
+    <#
+    .SYNOPSIS
+        Upgrade the Ed-Fi ODS/API AdminApp application in IIS.
+
+    .DESCRIPTION
+        Upgrades and configures the Ed-Fi ODS/API AdminApp application in IIS running in Windows 10 or
+        Windows Server 2016+. Admin App will be upgraded and reside under "Ed-Fi" website in IIS.
+        Appsettings values and connection strings will be copied over from existing Admin App application.
+        Invokes dbup migrations for updating the EdFi_Admin database accordingly.
+    .EXAMPLE
+        PS c:\> $parameters = @{
+            PackageVersion = '2.2.1'
+        }
+        PS c:\> Upgrade-EdFiOdsAdminApp @parameters
+
+        Upgrades Admin App to specified version and preseve the appsettings values and connection strings from existing application.
+    #>
+    [CmdletBinding(DefaultParameterSetName = 'Default')]
+    param (
+        # NuGet package name. Default: EdFi.ODS.AdminApp.Web.
+        [string]
+        $PackageName = "EdFi.Suite3.ODS.AdminApp.Web",
+
+        # NuGet package version. If not set, will retrieve the latest full release package.
+        [string]
+        $PackageVersion,
+
+        # NuGet package source. Defaults to "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json".
+        [string]
+        $PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json",
+
+        # Path for storing installation tools, e.g. nuget.exe. Default: "./tools".
+        [string]
+        $ToolsPath = "C:\temp\tools",
+
+        # Path for storing downloaded packages
+        [string]
+        $DownloadPath = "C:\temp\downloads",
+
+        # Path for the IIS WebSite. Default: c:\inetpub\Ed-Fi.
+        [string]
+        $WebSitePath = "c:\inetpub\Ed-Fi",
+
+        # Web site name. Default: "Ed-Fi".
+        [string]
+        $WebsiteName = "Ed-Fi",
+
+        # Web site port number. Default: 443.
+        [int]
+        $WebSitePort = 443,
+
+        # Path for the web application. Default: "c:\inetpub\Ed-Fi\AdminApp".
+        [string]
+        $WebApplicationPath = "C:\inetpub\Ed-Fi\AdminApp",
+
+        # Web application name. Default: "AdminApp".
+        [string]
+        $WebApplicationName = "AdminApp",
+
+        # TLS certificiate thumbprint, optional. When not set, a self-signed certificate will be created.
+        [string]
+        $CertThumbprint,
+
+        # Install Credentials: User
+        [Parameter(ParameterSetName="InstallCredentials")]
+        [string]
+        $InstallCredentialsUser,
+
+        # Install Credentials: Password
+        [Parameter(ParameterSetName="InstallCredentials")]
+        [string]
+        $InstallCredentialsPassword,
+
+        # Install Credentials: UseIntegratedSecurity setting
+        [Parameter(ParameterSetName="InstallCredentials")]
+        [switch]
+        $InstallCredentialsUseIntegratedSecurity,
+
+        # The hashtable must include: Server, Engine (SqlServer or PostgreSQL), and
+        # either UseIntegratedSecurity or Username and Password (Password can be skipped
+        # for PostgreSQL when using pgconf file). Optionally can include Port and
+        # DatabaseName.
+        [hashtable]
+        [Parameter(Mandatory=$true, ParameterSetName="InstallCredentials")]
+        $AdminDbConnectionInfo,
+
+        # Database Config
+        [switch]
+        $NoDuration
+    )
+
+    Write-InvocationInfo $MyInvocation
+
+    Clear-Error
+
+    $result = @()
+
+    $Config = @{
+        WebApplicationPath = $WebApplicationPath
+        PackageName = $PackageName
+        PackageVersion = $PackageVersion
+        PackageSource = $PackageSource
+        DatabaseInstallCredentials = @{
+            DatabaseUser = $InstallCredentialsUser
+            DatabasePassword = $InstallCredentialsPassword
+            UseIntegratedSecurity = $InstallCredentialsUseIntegratedSecurity
+        }
+        ToolsPath = $ToolsPath
+        DownloadPath = $DownloadPath
+        WebSitePath = $WebSitePath
+        WebSiteName = $WebsiteName
+        WebSitePort = $WebsitePort
+        CertThumbprint = $CertThumbprint
+        WebApplicationName = $WebApplicationName
+        NoDuration = $NoDuration
+        AppStartUp = $AppStartUp
+        ApplicationInstallType = "Upgrade"
+        AdminDbConnectionInfo = $AdminDbConnectionInfo
+    }
+
+    $elapsed = Use-StopWatch {
+        $result += Invoke-ApplicationUpgarde -Config $Config
+        $result += Get-AdminAppPackage -Config $Config
+        $result += Get-DbDeploy -Config $Config
+        $result += Invoke-TransferAppsettings -Config $Config
+        $result += Invoke-TransferConnectionStrings -Config $Config
+        $result += Install-Application -Config $Config
+        $result += Invoke-DbUpScripts -Config $Config
+
+        $result
+    }
+
+    Test-Error
+
+    if (-not $NoDuration) {
+        $result += New-TaskResult -name "-" -duration "-"
+        $result += New-TaskResult -name $MyInvocation.MyCommand.Name -duration $elapsed.format
+        $result | Format-Table
+    }
+}
+
+
 function Uninstall-EdFiOdsAdminApp {
     <#
     .SYNOPSIS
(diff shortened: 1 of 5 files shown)


EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -146,11 +146,11 @@ function Install-EdFiOdsAdminApp {
 
         # Path for storing installation tools, e.g. nuget.exe. Default: "./tools".
         [string]
-        $ToolsPath = "C:\temp\tools",
```
plioi, line 149: I am not asking for a change, but I think we originally started using C:\temp\ to avoid long folder path problems. Some of the packages we 'unzip' have amazingly-long paths, and nuget just quietly ignores those files, meaning we can get surprising side effect symptoms later when files are missing. Notice how the install.ps1 file still uses c:\temp, but then this and the upgrade script use $PSScriptRoot. That might also by why you needed to start gitignoring the 'local' tools folder.

What do you think about this concern of long paths? I think I ran into it recently in another context setting up my new laptop with a local ODS.

EdFi.Suite3.Installer.AdminApp/upgrade.ps1
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: Apache-2.0
+# Licensed to the Ed-Fi Alliance under one or more agreements.
+# The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+# See the LICENSE and NOTICES files in the project root for more information.
+
+import-module -force "$PSScriptRoot/Install-EdFiOdsAdminApp.psm1"
+
+<#
+Admin App will be upgraded to provided version. Appsettings values and connection strings will be copied over from existing Admin App application.
```
plioi, line 9: I'd phrase this like:

> This script will take your existing Admin App installation and upgrade it to the version indicated by `PackageVersion` below. Your existing appsettings.json config values and connection strings will be copied forward to the new version.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -324,6 +325,149 @@ function Install-EdFiOdsAdminApp {
...
+        [switch]
+        $InstallCredentialsUseIntegratedSecurity,
+
+        # The hashtable must include: Server, Engine (SqlServer or PostgreSQL), and
+        # either UseIntegratedSecurity or Username and Password (Password can be skipped
+        # for PostgreSQL when using pgconf file). Optionally can include Port and
+        # DatabaseName.
+        [hashtable]
+        [Parameter(Mandatory=$true, ParameterSetName="InstallCredentials")]
+        $AdminDbConnectionInfo,
```
plioi, line 413: This is marked as `Mandatory`, but upgrade.ps1 doesn't seem to have to provide it. Am I misinterpreting the meaning of when `Mandatory` really gets enforced?

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+                 # In case of existingApplicationPath is empty, then generating application physical path by combining site physical path and application path
+                $sitePath = $webSite.PhysicalPath
+                $appPath = $existingAdminAppApplication.path.trimstart('/')
+                $existingApplicationPath = "$sitePath\$appPath"
+            }
+            $versionString = [System.Diagnostics.FileVersionInfo]::GetVersionInfo("$existingApplicationPath\EdFi.Ods.AdminApp.Web.exe").FileVersion
+            $isVersionCompatible = IsExistingApplicationVersionCompatible $versionString
+            if($isVersionCompatible)
+            {
+                Write-Host "Found compatible version($versionString) of Admin App exists." -ForegroundColor Green
```
plioi, line 577: "We found a preexisting Admin App $versionString installation. Most likely, you intended to use the *upgrade* script instead of this *install* script."

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+                $confirmation = Read-Host -Prompt "Please enter 'y' to continue the installation or enter 'n' to stop the installation and do the upgrade( Note: On upgrade - all the appsettings and connection string values will be copied over from existing application folder) using upgrade.ps1"
```
plioi, line 578: "Please enter 'y' to continue the installation process, or enter 'n' to stop the installation so that you can instead run the upgrade script. Note: Using the upgrade script, all the appsettings and database connection string values would be copied forward from the existing installation, so only enter 'y' to continue if you are sure this is not an upgrade."

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+                if($confirmation -ieq 'y')
+                {
+                    $appsettingsFile =  Join-Path $existingApplicationPath "appsettings.json"
+                    if(Test-Path -Path $appsettingsFile)
+                    {
+                        Write-Host "Copying the encryption key from existing appsettings.json file at $appsettingsFile"
```
plioi, line 584: "To ensure your existing ODS / API Key and Secret values will continue to work, your existing encryption key is being copied forward from the appsettings.json file at $appsettingsFile"

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+                }
+                else
+                {
+                    Write-Warning "Exiting the installation."
+                    exit
+                }
+            }
+            else
+            {
+                Write-Warning "Found that existing version($versionString) of Admin App is not compatible with 2.2 and later. Please refer https://techdocs.ed-fi.org/display/ADMIN/Upgrading+Admin+App+from+1.x+Line for setting up the newer version of AdminApp."
```
plioi, line 597: "We found a preexisting Admin App $versionString installation. That version cannot be automatically upgraded in-place by this script. Please refer to https://techdocs.ed-fi.org/display/ADMIN/Upgrading+Admin+App+from+1.x+Line for setting up the newer version of AdminApp."

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
CSR2017: @plioi 
Since this is on "InstallCredentials" parameter set, so if I provide value to any of the parameters on that parameter set, then it will become mandatory.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+        $Config
+    )
+    Invoke-Task -Name ($MyInvocation.MyCommand.Name) -Task {
+
+        $existingWebSiteName = $Config.WebsiteName
+        $webSite = get-website | where-object { $_.name -eq $existingWebSiteName }
+        if($null -eq $webSite)
+        {
+            Write-Warning "Unable to find $existingWebSiteName on IIS."
+            $customWebSiteName = Read-Host -Prompt "If Ed-Fi website hosted with custom name, please enter the custom website name"
```
plioi, line 618: "Ed-Fi applications are usually deployed in IIS underneath a 'Ed-Fi' website entry. If you previously installed with a custom name for that entry other than 'Ed-Fi', please enter that custom name"

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+            # In case of existingApplicationPath is empty, then generating application physical path by combining site physical path and application path
+            $appPath = $existingAdminApp.path.trimstart('/')
+            $existingApplicationPath = "$existingWebSitePath\$appPath"
+        }
+
+        $versionString = [System.Diagnostics.FileVersionInfo]::GetVersionInfo("$existingApplicationPath\EdFi.Ods.AdminApp.Web.exe").FileVersion
+        $isVersionCompatible = IsExistingApplicationVersionCompatible $versionString
+        if(-not $isVersionCompatible)
+        {
+            throw "Upgrade Admin app option not supported on existing version $versionString. Only supported from version 2.2+. Please refer https://techdocs.ed-fi.org/display/ADMIN/Upgrading+Admin+App+from+1.x+Line for setting up the newer version of AdminApp."
```
plioi, line 656: "We found a preexisting Admin App $versionString installation. That version cannot be automatically upgraded in-place by this script. Please refer to https://techdocs.ed-fi.org/display/ADMIN/Upgrading+Admin+App+from+1.x+Line for setting up the newer version of AdminApp."

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,247 @@ function Uninstall-EdFiOdsAdminApp {
...
+        [Parameter(Mandatory=$true)]
+        $Config
+    )
+
+    Invoke-Task -Name ($MyInvocation.MyCommand.Name) -Task {
+
+
+        $backUpPath = $Config.ApplicationBackupPath
+        Write-Warning "The following appsettings will be copied over from existing application: "
+        $appSettings = @('ProductionApiUrl','SecurityMetadataCacheTimeoutMinutes','DatabaseEngine', 'AppStartup', 'EncryptionKey', 'Log4NetConfigFileName')
```
plioi, line 735: Copying ApiStartup**Type** forward seems really important, too. ApiExternalUrl, XsdFolder, PathBase too.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
plioi: Cool, that makes sense. I'm marking this resolved with no need for a change.

plioi:
I approve of the changes except for the one open question about C:\temp vs $PSScriptRoot. As discussed, although $PSScriptRoot is wise 99% of the time, in this case we are trying to mitigate the risk of extremely-long folder paths inside our install-time nuget dependencies, which can cause some packages to "successfully" install while missing files that later cause trouble.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
stephenfuqua: This is a very good point about long file names, I've certainly run into that myself just this week and had to change my direction on a script from using `$PSScriptRoot` to something shorter. In my case (Engage Online Learners Starter Kit), I switched to a default value of `c:/Ed-Fi`.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,251 @@ function Uninstall-EdFiOdsAdminApp {
     }
 }
 
+
+function Invoke-InstallationPreCheck{
+    [CmdletBinding()]
+    param (
+        [hashtable]
```
stephenfuqua, line 553: I find it much harder to use functions that accept a `hashtable` instead of directly listing their arguments. That said, I don't want to hold this ticket up over that. If your sprint is ahead of schedule (that is, if you have time), please _consider_ changing this. Only asking for consideration, not immediate agreement.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -402,6 +546,251 @@ function Uninstall-EdFiOdsAdminApp {
...
+        $existingApplicationPath = ($existingAdminApp).PhysicalPath
+        if(!$existingApplicationPath)
+        {
+            # In case of existingApplicationPath is empty, then generating application physical path by combining site physical path and application path
+            $appPath = $existingAdminApp.path.trimstart('/')
+            $existingApplicationPath = "$existingWebSitePath\$appPath"
+        }
+
+        $versionString = [System.Diagnostics.FileVersionInfo]::GetVersionInfo("$existingApplicationPath\EdFi.Ods.AdminApp.Web.exe").FileVersion
+        $isVersionCompatible = IsExistingApplicationVersionCompatible $versionString
```
stephenfuqua, line 653: I see some code duplication here that could be refactored.

However, recently the newest member of the ODS Platform team introduced me to the acronym WET as the opposite of DRY: Write Everything Twice. As I understood it, the point was being made that simply copying and pasting this was really easy and allowed you to quickly move on with the coding. Once you need it a third time though, then it is worth the effort.

There have been a few times where I thought that was worth while, but I'm not completely convinced. This function is much longer than I like to read all at once, so even if not for the DRY principle, I would prefer to read this block of code as something like `$compatibleVersionIsInstalled = CheckForCompatibleVersion($existingAdminApp)`.

Now, I did not look very closely. There might be more code above this that could be incorporated into such a function.

What do you think - will it be worth it to refactor a little bit to (a) reduce length of the function, (b) reduce duplication, and (c) simplify the process of reading and understanding what this function is trying to achieve?

Again, this question is completely at your discretion, and I'm not going to hold up the ticket over it.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -499,9 +888,15 @@ function Invoke-TransformAppSettings {
         $settings.AppSettings.SecurityMetadataCacheTimeoutMinutes = '10'
         $settings.AppSettings.DatabaseEngine = $config.engine
         $settings.AppSettings.AppStartup = $Config.AppStartUp
+
         if(!$settings.AppSettings.EncryptionKey)
         {
-            $settings.AppSettings.EncryptionKey = New-AESKey
+            $encryptionKey = New-AESKey
```
stephenfuqua, line 894: How long does it take to run the `New-AESKey` on a low-resource virtual machine? Would it be worthwhile to change this around so that it is in an `else` clause below instead? I don't remember if this is a super trivial function or one of those compute intensive encryption functions.

EdFi.Suite3.Installer.AdminApp/Install-EdFiOdsAdminApp.psm1
```diff
@@ -144,13 +144,13 @@ function Install-EdFiOdsAdminApp {
         [string]
         $PackageSource = "https://pkgs.dev.azure.com/ed-fi-alliance/Ed-Fi-Alliance-OSS/_packaging/EdFi/nuget/v3/index.json",
 
-        # Path for storing installation tools, e.g. nuget.exe. Default: "./tools".
+        # Path for storing installation tools, e.g. nuget.exe. Default: "C:\temp\tools".
         [string]
-        $ToolsPath = "$PSScriptRoot/tools",
+        $ToolsPath = "C:\temp\tools",
```
stephenfuqua, line 149: 👍

plioi (approved):

PR #214, opened 2021-09-24
[AA-1361] Show Error Message on Upgrade Script When Version Specified Is Same or Previous

pmcvtm:
Adds a new check in the Upgrade script to verify the specified upgrade version is newer than the existing install.

In addition:

- adds helper function for version checking
  - update other compatibility check to use it as well
- minor readability refactors (favor early exit over `if/else`


pmcvtm:
Added new test cases for upgrade messages and changed the "install" script behavior when an existing Admin App exists to always exit. Depending on the target and existing versions, the script will point the user towards the upgrade script, ask them to manually upgrade, or ask them to uninstall / reinstall for "downgrading."

pmcvtm:
- Restored the prompt when new version is OK to be `install`ed over previous, with recommendation to use `upgrade` instead.
  - Changed message for emphasis
- Added missing test case for new "Install downgrade" case

CSR2017 (approved):

PR #215, opened 2021-09-30
[AA-1373] Add a global error handler to Admin App

CSR2017:
@saa14 @pmcvtm 
Just added AdminAppException type, we can update or manipulate as needed.
Thanks


Application/EdFi.Ods.AdminApp.Web/ErrorHandler/GlobalErrorHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+                adminAppError.Status = webException.Status.ToString();
+                adminAppError.InnerException = webException.InnerException;
+
+            }
+            catch (Exception exception)
+            {
+                _logger.Error(exception);
+                adminAppError.StackTrace = exception.StackTrace;
+                adminAppError.ErrorMessage = exception.Message;
+                adminAppError.InnerException = exception.InnerException;
```
stephenfuqua, line 40: @CSR2017 this object's values are set, but the object itself is never used.... ?

But happy to see a global error handler being added!

CSR2017: @stephenfuqua 
Thanks for the feedback. Just creating and filling AdminAppException object. This will be used by upcoming code implementation (https://tracker.ed-fi.org/browse/AA-1377). If it looks orphan object, will remove it. Thanks

stephenfuqua: I would add a comment then, so that anyone looking at the `main` branch after this merges will understand that this is not dead code - it is building up to another feature. So I would just put something like `// TODO: in AA-1377 we will start making use of this adminAppError object`.

pmcvtm:
We talked offline about this but want to capture here: There are some inconsistencies with the flow due to the `HandleAjaxErrorAttribute` taking precedence and the `GlobalErrorHandler` not processing the errors to the actual response.

We discussed that we _could_ "re-order" the operations, cut down what the Ajax handler does, and re-`throw;`ing exceptions in the global handlers so they "pass through" to the next handler.

But it would be simpler to consolidate the handling into the global error handler:
- move the AJAX check and response generation from `HandleAjaxErrorAttribute` into `GlobalErrorHandler` and then remove `HandleAjaxErrorAttribute`
- then in `GlobalErrorHandler` when the request is not AJAX, redirect to the error page (`Home/Error`)

Application/EdFi.Ods.AdminApp.Web/ErrorHandler/GlobalErrorHandler.cs
```diff
@@ -0,0 +1,68 @@
...
+                _logger.Error(exception);
+
+                if (context.Request.IsAjaxRequest())
+                {
+                    context.Response.StatusCode = (int)HttpStatusCode.InternalServerError;
+                    var controllerName = context.Request.RouteValues["controller"].ToString();
+                    var responseText = IsReportsController(controllerName) && exception is SqlException
+                        ? "An error occurred trying to access the SQL views for reports."
+                        : exception.Message;
+                    context.Response.WriteAsync(responseText);
```
saa14, line 43: Since `WriteAsync` is asynchronous, this call needs to be awaited. Consider adding an `await` to this call and correspondingly converting HandleException to an async function which also needs to be awaited at line 29

saa14 (approved):

PR #217, opened 2021-10-07
[AA-1378] Support for multiple database servers in Admin App

CSR2017:
@saa14 @pmcvtm 
Thanks

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Management/Services/ConnectionStringBuilderAdapterFactory.cs
+++ b/Application/EdFi.Ods.AdminApp.Management/Services/ConnectionStringBuilderAdapterFactory.cs
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using Microsoft.Extensions.Options;
+using System;
+using EdFi.Ods.AdminApp.Management.Helpers;
+using EdFi.Common.Database;
+using EdFi.Common.Configuration;
+
+namespace EdFi.Ods.AdminApp.Management.Services
+{
+    public interface IConnectionStringBuilderAdapterFactory
+    {
+        IDbConnectionStringBuilderAdapter Get();
+    }
+
+    public class ConnectionStringBuilderAdapterFactory : IConnectionStringBuilderAdapterFactory
+    {
+        private IOptions<AppSettings> _appSettings;
+
+        public ConnectionStringBuilderAdapterFactory(IOptions<AppSettings> appSettings) => _appSettings = appSettings;
+
+        public IDbConnectionStringBuilderAdapter Get()
+        {
+            var isPostgreSql = ApiConfigurationConstants.PostgreSQL.Equals(_appSettings.Value.DatabaseEngine,
+            StringComparison.InvariantCultureIgnoreCase);
+
+            if (isPostgreSql)
+            {
+                return new NpgsqlConnectionStringBuilderAdapter();
+            }
+            return new SqlConnectionStringBuilderAdapter();
+        }
+    }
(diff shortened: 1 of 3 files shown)


Application/EdFi.Ods.AdminApp.Management/Services/ConnectionStringBuilderAdapterFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using Microsoft.Extensions.Options;
+using System;
+using EdFi.Ods.AdminApp.Management.Helpers;
+using EdFi.Common.Database;
+using EdFi.Common.Configuration;
+
+namespace EdFi.Ods.AdminApp.Management.Services
+{
+    public interface IConnectionStringBuilderAdapterFactory
```
pmcvtm, line 14: I am not sure I understand the value of the factory here, since we're not stubbing it in different tests, or doing anything specific / atypical with how we register or implement these interfaces.

Can we move this logic to `ConnectionStringService`?

Application/EdFi.Ods.AdminApp.Management/Services/ConnectionStringBuilderAdapterFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        IDbConnectionStringBuilderAdapter Get();
+    }
+
+    public class ConnectionStringBuilderAdapterFactory : IConnectionStringBuilderAdapterFactory
+    {
+        private IOptions<AppSettings> _appSettings;
```
pmcvtm, line 21: This should be made `readonly` if we keep it (or when it moves to `ConnectionStringService`)

Application/EdFi.Ods.AdminApp.Management/Services/ConnectionStringService.cs
```diff
@@ -7,50 +7,67 @@
...
                 {
-                    connectionString = GetModifiedConnectionString(
-                        connectionString, $"Ods_{odsInstanceName.ExtractNumericInstanceSuffix()}");
+                    if (IsTemplate(input))
+                    {
+                        updatedValue = GetModifiedString(input,$"Ods_{odsInstanceName.ExtractNumericInstanceSuffix()}");
+                    }
+                    else
+                    {
+                        if(isDbCheck)
```
pmcvtm, line 47: We might do the `isDbCheck` outside of the function, or in a different function, to remove that parameter / check and keep this function strictly "replace" -oriented
```
public string GetConnectionString(string odsInstanceName, ApiMode apiMode)
{
    var connectionStringBuilder = _connectionStringBuilderAdapterFactory.Get();
    connectionStringBuilder.ConnectionString = _connectionStrings.ProductionOds;
    
    if(apiMode.SupportsMultipleInstances && !IsTemplate(connectionStringBuilder.DatabaseName)
        throw new InvalidOperationException(...);

    connectionString = GetModifiedConnectionString(connectionString);

    string GetModifiedConnectionString(string connectionString)
       //check mode and replace template as currently
}
```

Application/EdFi.Ods.AdminApp.Management/Services/ConnectionStringService.cs
```diff
@@ -7,50 +7,67 @@
...
-            if (apiMode.SupportsMultipleInstances)
+            string GetUpdatedName(string input, bool isDbCheck)
             {
-                if (IsTemplate(connectionString))
+                var updatedValue = input;
+                if (apiMode.SupportsMultipleInstances)
                 {
-                    connectionString = GetModifiedConnectionString(
-                        connectionString, $"Ods_{odsInstanceName.ExtractNumericInstanceSuffix()}");
+                    if (IsTemplate(input))
```
pmcvtm, line 41: It's [not necessary to check](https://stackoverflow.com/questions/34404042/should-i-use-string-contains-before-string-replace) for `contains` before calling `replace` (nor is it an optimization).

We can remove this (and the one on line 56) to cut down on the number of `if`/`else`s in this function.

PR #219, opened 2021-10-19
[AA-1234] - Learning Standards credentials check during form submission

saa14:
**Description**

**Controller and Validator changes**

- Added SafeCustomAsync similar to SafeCustom for async validation rules. 
- Added the validation rule to validate api credentials in the LearningStandardsModelValidator. This borrows from the validation check in LearningStandardsJob.cs (which will be removed once we have this move completed). Added LearningStandardsViewModel to capture only the ApiKey and ApiSecret properties from the LearningStandards form.
- As mentioned in the FluentValidation docs (https://docs.fluentvalidation.net/en/latest/async.html), since we are using async validation rules, it's important to call ValidateAsync on the validator. LearningStandardsViewModel is used to capture the form input and allow us to call ValidateAsync ourselves instead of allowing automatic handling by FluentValidation.
- Refactored the LearningStandards controller and action to validate the LearningStandards credentials using FluentValidator's recommended ValidateAsync approach. Returned a JsonResult with validation errors to comply with the site-form-handlers.js functions.
- Removed the LearningStandards credentials validation check and the corresponding method from the RunJob method as the validation check is now moved to the LearningStandardsModelValidator.

**UI changes**

- Separated out enablingLearningStandards and updateLearningStandards for the Enable and Update Learning Standards forms. enablingLearningStandards uses the site-form-handlers.js functions to make the ajax request. Added loading spinner to the enabling Learning Standards button.

**Test changes**

- Updated the LearningStandards mock controller with the mock for the LearningStandardsModelValidator.
- Refactored to use the updated model type (LearningStandardsViewModel) for the LearningStandards action.


pmcvtm (commented):
Looks good and tests well, see other comments for some minor cleanup to do before merging.

PR #224, opened 2021-10-26
[AA-1375] - Added UI for error handler

saa14:
**Description**
- Added ErrorModel for the Error screens
- Moved the condition-based view components from the Error view to a partial in ErrorFeedback so that it can be used for AJAX errors as well.
- Refactored Error action in the Home Controller to use the Error Model and partial view for AJAX Errors.
- Refactored the LoadAsyncAction function to load the partial view result in case of Ajax errors. Moved the additional error message from the AjaxAction HtmlHelperExtension to the bearerTokenResponse errorMessage. Refactored the AjaxAction HtmlHelperExtension to remove the not-needed data attributes.
- Refactored the FieldValues() function to capture all the details (stack trace and status) from the Error view.
- Removed unused logger field from the HomeController as it was moved to the ErrorModel. Corrected the type for the logger in the ErrorModel.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Management/Api/TokenRetriever.cs
+++ b/Application/EdFi.Ods.AdminApp.Management/Api/TokenRetriever.cs
@@ -59,9 +59,7 @@ namespace EdFi.Ods.AdminApp.Management.Api
 
             var bearerTokenResponse = oauthClient.Execute<BearerTokenResponse>(bearerTokenRequest);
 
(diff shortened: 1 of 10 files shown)


pmcvtm (approved):
These changes look great; we're covering all our bases now between regular / AJAX requests, and while the issue collector handling is "interesting" it makes for a consistent experience. Thanks!

PR #227, opened 2021-11-02
[AA-1384] Find a method of referring to ODS Instances that allows longer then 50 character

CSR2017:
@saa14 @pmcvtm 
1. Updated Ods instance registration model to have new DatabaseName column for holding the associated db name
2. Update the instance to hold just instance id (district id or year)
3. Refactored model validation 
4. Added sql and postgresql migration scripts
Thanks


pmcvtm (approved):
This looks good, and I tested with a couple configurations. Nice bonus that this cuts down on some of the DB-checking complexity.

PR #230, opened 2021-11-04
[AA-1380] Admin App Installer issue on re-installation 

CSR2017:
@pmcvtm @saa14 
1. Removed WebApplicationDirectory, since it was confusing to have both WebApplicationname and WebApplicationDirectory. 
If user wants to provide custom name for AdminApp application, it can be provided with WebApplicationname  parameter. Internally the WebApplication path will be generated by combining Websitepath and WebApplicationname
2. Actual error (AA-1380) cannot be reproduced, since user doesn't have option to provide two different values for WebApplicationname and WebApplicationDirectory as before.
3. Testing: General testing on install, upgrade and uninstall flows. For testing with custom values Install-WithCustomSettings, Upgrade-ApplicationWithCustomSettings scenarios from .\test.ps1 can be used.
Thanks


pmcvtm (approved):
Changes look good and functionality and tests work as expected!

PR #231, opened 2021-11-05
[AA-1392] Success Message When Delete Fails

pmcvtm:
Fix the ODS API client `Delete` handler to check if the response is a success and set the error message appropriately (instead of just on exception).

Included default message as well as potentially a Content message. Other client methods are correctly checking this and changing the status accordingly.

This _does not_ adjust the delete JavaScript handling to wire into the new error page handling, so it still shows a toast.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Management.UnitTests/Api/OdsApiFacadeTests.cs
+++ b/Application/EdFi.Ods.AdminApp.Management.UnitTests/Api/OdsApiFacadeTests.cs
@@ -312,13 +312,11 @@ namespace EdFi.Ods.AdminApp.Management.UnitTests.Api
             _facade = new OdsApiFacade(_mapper, mockOdsRestClient);
 
             // Act
-            var ex = Assert.Throws<OdsApiConnectionException>(() => _facade.GetAllDescriptors());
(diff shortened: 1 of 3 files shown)


Application/EdFi.Ods.AdminApp.Management.UnitTests/Api/OdsApiFacadeTests.cs
```diff
@@ -292,6 +293,34 @@ public void Should_DeleteSchool_returns_not_success_result_when_api_throws_excep
...
+            var mockTokenRetriever = new Mock<ITokenRetriever>();
+            mockTokenRetriever.Setup(x => x.ObtainNewBearerToken()).Returns("Token");
+
+            var mockOdsRestClient =
+                new OdsRestClient(_connectionInformation, mockRestClient.Object, mockTokenRetriever.Object);
+
+            _facade = new OdsApiFacade(_mapper, mockOdsRestClient);
+
+            // Act
+            var ex = Assert.Throws<OdsApiConnectionException>(() => _facade.GetAllDescriptors());
```
CSR2017, line 315: @pmcvtm 
I guess, we may need to use DeleteSchool method here.

CSR2017 (approved):

PR #240, opened 2021-11-18
[AA-1358] - Add support for Docker Health checks

saa14:
**Description**
- Added AspNetCoreHealthCheck packages to perform the database probes needed for Postgres and SQLServer admin database during the docker healthcheck.
- Added HealthCheckServiceExtensions class to add the database health check depending on the database engine being used. Added IsSqlServer method to Startup to check for the database engine and refactored using Staertup.cs using this check. Called MapHealthChecks on the endpoint builder with the health check endpoint.


pmcvtm (approved):
Looks good and confirmed healthchecks work as expected. Nice find for that library. There were a couple minor whitespace issues I cleaned up.

PR #245, opened 2021-12-20
[AA-1405] Initial GitHub action

andonyns:
Creating initial GitHub action that includes some of the steps of the existing TeamCity build:

- Build Admin App web
- Populate Google Analytics ID
- Run Unit tests.

This needed a dot net restore since the restore included in the build script is not working on the GitHub Action


.github/workflows/build.yml
```diff
@@ -0,0 +1,38 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
pmcvtm, line 7: Let's remove this branch stipulation so we can trigger builds for _any_ pull request.
I know they're not common but we still want to run builds for PRs that will merge into "epic" branches.

The syntax looks a little weird but I think it will end up being to remove the filter:
```yaml
pull_request:

env:
# ...
```

pmcvtm (approved):
Approving this but holding off on merge until @stephenfuqua has a chance to apply the necessary secrets, to avoid adding new unfixable :x:es on our other PRs.

.github/workflows/build.yml
pmcvtm: For anyone reading this in the future, GH must think this syntax is weird too, as they call it out in [the docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration):
> You must append a colon (:) to all events, including events without configuration.

pmcvtm:
We're green! ✔️ Merging.

PR #250, opened 2022-01-12
[ODST-194] Run playwright tests in action

andonyns:
**This should be merged after https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/pull/248 is merged**

Run playwright tests against the ODS instance created.

- The purpose of these task is not for the tests to pass, just to run. The new tests will be added in a future ticket.
- The results are currently not saved since this will be covered in a future ticket.

Needs https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Docker/pull/102 to be merged to fix issue with case of folder names

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  setup-ods:
+  run-e2e-tests:
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e.yml
```diff
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  setup-ods:
+  run-e2e-tests:
```
pmcvtm, line 15: A note for posterity: In deciding to stick with a single job, we learned that unlike the very similar Azure DevOps, GitHub actions [does not have "special" syntax for single-job actions](https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-azure-pipelines-to-github-actions#key-differences), and you must always declare a job.

pmcvtm (approved):
This looks good! To re-iterate for future viewers: the goal here was to _run_ end-to-end tests against the ODS docker container, not for them to be passing. Excited to see where this is heading!

PR #260, opened 2022-04-08
[AA-1423] - Upgrade to dotnet 6

saa14:
**Description**
**Upgrades/Migration**
- Updated the target framework to netstandard 2.1 for the Management projects.
- Updated the target framework to net6.0 for Azure.IntegrationTests and Management.Tests projects since they have a dependency on AdminApp.Web project.
- Updated the target framework to net6.0 for the EdFi.Ods.AdminApp.Web project.
- Upgraded nuget packages on AdminApp.Web to the versions compatible with net6.0.
- Upgraded nuget packages on AdminApp.Management to the versions compatible with netstandard2.1.Upgraded NUnit nuget packages on the Management projects to the versions compatible with net6.0/netstandard2.1.
- Updated global.json sdk version to net6.0.
- Updated github action workflow to use .net 6.0.x instead of .net core 3.1.x.Updated developer docs to refer to .net 6.0 instead of .net core 3.1.

**Refactoring Work after package upgrades**
- Refactored ViewModels and ValidationExtensions to use ValidationContext<T> instead of Custom Context and IRuleBuilderOptionsConditions instead of IRuleBuilderInitial after the version upgrade to FluentValidation packages.
- Refactored to add configuration details as init-only properties instead of using setters (which are now removed) after upgrading to the latest CsvHelper version.
- Refactored to resolve warnings after nuget package upgrades.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.x
+        dotnet-version: 6.0.x
     - name: Restore dependencies
       run: dotnet restore
       working-directory: ./Application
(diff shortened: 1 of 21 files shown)


pmcvtm (approved):
This looks great! I am _slightly_ disappointed we are not able to target the `Management` projects at a "library" framework version as we prepare for things to come, but more important that things work.

Posterity note: We are immediately following this with work to resolve JS/CSS bundling issues as well as upgrade other dependencies before the next release.

PR #261, opened 2022-04-14
[AA-1479] - Replace BuildBundlerMinifier package with WebOptimizer on Admin App as we upgrade to .NET 6

saa14:
**NOTE: This PR should be merged only after [the .NET 6 upgrade PR](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/pull/260) has been merged**

**Description**
- Added WebOptimizer package to the Web project.
- Setup WebOptimizer in the Startup and added the bundle/minification configurations similar to bundleconfig.json. Removed bundleconfig.json.
- Added tag helpers from WebOptimizer.Core.


pmcvtm (approved):
Looks good! Parity with bundling between systems and the styles and scripts are loading both for local / IDE run as well as for the installer.

PR #264, opened 2022-04-25
AA-1460 - Remove Azure Support from Admin App

saa14:
**Description**
- Removed AdminApp.Management.Azure project.Removed AdminApp.Management.Azure.UnitTests project.Removed AdminApp.Management.Azure.IntegrationTests project.Removed LegacyNET48Web App Files folder.
- Removed Azure related classes in AdminApp.Web project.Removed Azure related classes in AdminApp.Management project.
- Removed GetSqlConfiguration() from IOdsSecretConfigurationProvider. Removed corresponding implementations/functions for the removed AzureSqlConfigurationRefactored OdsSecretConfigurationProviderTests to remove AzureSqlConfiguration.Removed unused AzureSqlConfigurations property from AdminAppDbContext.cs
- Refactored FirstTimeSetup tests to use CompleteOnPremFirstTimeSetupCommand and OnPremFirstTimeSetupService instead of the removed Azure counterparts.
- Added migrations to remove the adminapp.AzureSqlConfigurations table if it exists.
- Updated Web.csproj after the Azure management project was removed.Updated Ed-Fi-ODS-Tools.sln after Azure related projects were removed.


pmcvtm:
Two small changes we should make before merging:

- [ ] Let's revert the VS version change (lines 3+4) in `Application/Ed-Fi-ODS-Tools.sln` to avoid locking out folks who haven't upgraded yet
  - We might elect to change this closer to or after the next release, since we _do_ want to keep up with the times.
- [ ] It looks like the main `appsettings.json` has "Azure" settings as defaults. Even though we typically run with the environmental transform settings, we should replace those so the main one stays accurate/relevant.

Another note: it looks like CSS styles are not loading. I know we just addressed that in #261 so check to makes sure you're merged up with main and/or rebased correctly.

pmcvtm (approved):
Looks great, thanks for making those changes!

PR #265, opened 2022-04-26
[AA-1457] Separate YML files into different files to allow test reports to be generated in forks

andonyns:


.github/workflows/zephyr-results.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches: [ main ]
+
+env:
+  JIRA_ACCESS_TOKEN: ${{ secrets.JIRA_ACCESS_TOKEN }}
+  ADMIN_APP_VERSION: '2.3.1'
+  PROJECT_ID: '11900'
+  CYCLE_NAME: 'Automation Cycle'
+  TASK_NAME: 'Playwright Automation Task'
+  FOLDER_NAME: 'Playwright Automation Run'
+
```
pmcvtm, line 16: Are we missing the `RESULTS_FILE` variable from the previous workflow here? Looks like it's still in use below.

andonyns: Thanks! I will correct this

pmcvtm (approved):

PR #270, opened 2022-04-29
[AA-1504] Improvements to script to send result to Zephyr

andonyns:
- Move the script to the eng folder
- Added logging for more information
- Fix Admin App version
- Throw error if specified version is not found (This was currently adding the results to an unscheduled release, and this will make the results get lost into the executions)


pmcvtm (approved):
Looks great. Big improvement in information provided + error handling / flow. Any future issues or changes to how this work will be much clearer in what is happening!

PR #276, opened 2022-05-18
[AA-1506] Claim set editor - claim set import not retaining the authorization strategy overrides (3.0)

CSR2017:
@pmcvtm 
Thanks


pmcvtm (approved):
Looks good!

I notice a lot of verbosity / complexity around usages `AuthorizationStrategy[]`, which you did a good job of extracting out to make more readable in this new code. I wonder if it'd be worth a tech-debt ticket in the future to implement a custom collection type to make that even smoother and more readable.

PR #278, opened 2022-05-18
[AA-1462] Admin API Authentication

pmcvtm:
Introduce authentication and authorization to Admin API using the [OpenIddict](https://documentation.openiddict.com/index.html) library.

- follows "client credential" flow to serve auth tokens for use by other applications
  - supports a single "full access" scope that is granted and required on all endpoints
- registration is disabled by default in prod, but overridden when no Applications are registered
- when testing, you can [disable token encryption](https://documentation.openiddict.com/configuration/token-formats.html#disabling-jwt-access-token-encryption) to validate contents in service such as [jwt.io](https://jwt.io/)

**Remarks**

- OpenIddict endpoints do not return responses in the same format as the rest of AdminAPI, which can be expected as it's its own 'server' context
- registers `/connect/` endpoints as traditional MVC Controllers in order to support `form-urlencoded` requests (per OAuth2) and to work around odd `jwt` serialization issue
  - Controller registration is "obscured" away in `SecurityExtensions` to discourage other controllers
- uses custom entity types (in separate schema) for naming + in case we move away from OpenIddict in the future
- OpenIddict gleans request from `HttpContext`, so the `/connect/token` endpoint does not include a body. The schema is added manually to swagger, although the "Try It Out" button does not work (as noted in description). The "Authorize" button and Postman work as expected

### DIFF
--- a/Application/EdFi.Ods.Admin.Api/Infrastructure/Security/TokenEndpointBodyDescriptionFilter.cs
+++ b/Application/EdFi.Ods.Admin.Api/Infrastructure/Security/TokenEndpointBodyDescriptionFilter.cs
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace EdFi.Ods.Admin.Api.Infrastructure.Security;
+
+/// <summary>
+/// Swashbuckle OperationFilter to manually specify the Request Body for the Token endpoint,
+/// which pulls the request from HttpContext as per OpenIddict convention.
+/// </summary>
+public class TokenEndpointBodyDescriptionFilter : IOperationFilter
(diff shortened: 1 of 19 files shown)


Application/EdFi.Ods.Admin.Api/Infrastructure/Security/TokenEndpointBodyDescriptionFilter.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    private OpenApiMediaType BuildTokenRequestBodyDescription() => new OpenApiMediaType
+    {
+        Schema = new OpenApiSchema
+        {
+            Type = "object",
+            Properties = new Dictionary<string, OpenApiSchema>
+            {
+                { "client_id", new OpenApiSchema { Type = "string "} },
+                { "client_name", new OpenApiSchema { Type = "string "} },
```
CSR2017, line 40: client_name parameter is not accurate. Needs to be updated to "client_secret".

CSR2017, line 40: client_name parameter is not accurate. Needs to be updated to "client_secret".

pmcvtm: Great catch! It looks like that 'fixes' this endpoint using the Try It Out button, too, so I clarified the message that's appended to the description as well.

CSR2017 (approved):

PR #281, opened 2022-05-30
[AA-1406] Getting Admin App version dynamically form Azure Artifacts

andonyns:
- Creating script to get the version
- Adding compose file that uses published ODS/API images and builds Admin App locally
- Fixing navigation issue when website is not up


pmcvtm (approved):
Looks good. Discussed need for manual upkeep here and possible alternatives if that becomes unwieldy.

PR #282, opened 2022-06-01
[AA-1463] Add Vendors API endpoint to Admin API

CSR2017:
@pmcvtm 
Please help review and merge the changes


Application/EdFi.Ods.AdminApp.Management/Database/Commands/EditVendorCommand.cs
```diff
@@ -20,9 +23,18 @@ public EditVendorCommand(IUsersContext context)
             _context = context;
         }
 
-        public void Execute(IEditVendor changedVendorData)
+        public Vendor Execute(IEditVendor changedVendorData)
         {
             var vendor = _context.Vendors.Single(v => v.VendorId == changedVendorData.VendorId);
```
pmcvtm, line 28: MUST: Change to `SingleOrDefault(...)`

The below code null-checking code will not execute since `Single` will throw an `InvalidOperationException` when the result is not found, or matches more than one result.

Application/EdFi.Ods.AdminApp.Management/Database/Commands/EditVendorCommand.cs
```diff
@@ -20,9 +23,18 @@ public EditVendorCommand(IUsersContext context)
...
+            if(vendor == null)
+            {
+                throw new NotFoundException<int>("This vendor no longer exists.It may have been recently deleted.", changedVendorData.VendorId);
```
pmcvtm, line 31: MUST: Change message to simply `"vendor"` instead of a full sentence.

`NotFoundException` builds up a similar message and takes only `resourceName` as a parameter. If you'd like you can edit the details of that message to match this instead, or catch the `NotFoundException` and return a different message in the controller.

pmcvtm: Please also make this fix in `VendorFeature.GetVendor()` and `DeleteVendorCommand.Execute()`

pmcvtm:
SHOULD: Let's rearrange our "Feature" code to really make use of that feature-slice pattern we're using.

Instead of following an MVC-ish Models/Endpoints pattern where we have classes `AddXModel.cs`, `EditXModel.cs` and `XFeature.cs` with `Add()` and `Edit()` methods, we can break each feature into its own containing class: `AddX.cs` and `EditX.cs` with sub-classes and methods to match. Organizing our code this way isolates and groups the components of each endpoint to the point that it breaks pattern to cross between features.

You can see this in practice with the `TokenService` and `RegisterService` without any endpoint registration, but here is an example of what `AddX.cs` might look like:

```csharp
public class AddX : IFeature
{
    public void MapEndpoints(IEndpointRouteBuilder endpoints)
    {
        endpoints.MapPost("/x", Handle).RequireAuthorization();
    } 

    public class Request
    {
        //properties
    }

    public async Task<IResult> Handle(Request request, IAddXService service)
    {
        //functionality to add X
    }

    public class Validator : AbstractValidator<Request>
    {
        public Validator()
        {
            // validates new X can be added
        }
    }
}
```

pmcvtm (commented):
This looks pretty good! Still working through this but wanted to share what I've found so far. Please see comment on the main thread about code structure as well.

stephenfuqua:
`IFeature` - a pattern I've read about but never tried. Look forward to seeing how it works out.

Application/EdFi.Ods.Admin.Api/Features/Vendors/EditVendor.cs
```diff
@@ -0,0 +1,55 @@
...
+using EdFi.Ods.AdminApp.Management.Database.Queries;
+using FluentValidation;
+
+namespace EdFi.Ods.Admin.Api.Features.Vendors
+{
+    public class EditVendor : IFeature
+    {
+        public void MapEndpoints(IEndpointRouteBuilder endpoints)
+        {
+            endpoints.MapPut($"/{FeatureConstants.Vendors}", Handle).RequireAuthorization()
```
andonyns, line 18: Consider adding URL to be /id as done in the DELETE and GET individual vendor

pmcvtm: Good catch, this would match our intended design as well.

Application/EdFi.Ods.Admin.Api/Infrastructure/WebApplicationBuilderExtensions.cs
```diff
@@ -1,19 +1,64 @@
...
 
 public static class WebApplicationBuilderExtensions
 {
     public static void AddServices(this WebApplicationBuilder webApplicationBuilder)
     {
+        var executingAssembly = Assembly.GetExecutingAssembly();
+        webApplicationBuilder.Services.AddAutoMapper(executingAssembly);
+        webApplicationBuilder.Services.AddScoped<InstanceContext>();
+
+        foreach (var type in typeof(IMarkerForEdFiOdsAdminAppManagement).Assembly.GetTypes())
```
pmcvtm, line 23: COMMENT: This is great. I know it was some extra work up-front but this will definitely save us some headaches in the future.

Application/EdFi.Ods.Admin.Api/ActionFilters/OrderOperationsDocumentFilter.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed to the Ed-Fi Alliance under one or more agreements.
+// The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
+// See the LICENSE and NOTICES files in the project root for more information.
+
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace EdFi.Ods.Admin.Api.ActionFilters
+{
+    public class OrderOperationsDocumentFilter : IDocumentFilter
```
pmcvtm, line 11: Good call on getting ahead of adding this. I did a little cleanup to "explain" some of that lengthy LINQ logic and resolved the nullable warnings below.

pmcvtm (approved):
Looks great! Thanks for re-organizing and fixing up the Swagger doc in the aftermath. I added on a few small commits to resolve the nullability warnings and fix a (unrelated?) issue preventing me from launching the regular Admin Web app from Rider.

PR #285, opened 2022-06-13
[AA-1468] Adding GitHub Action setup to publish Admin Api packages

andonyns:
- Adding method to publish Admin API to Azure Artifacts
- Adding method to promote to release for Admin API, AdminApp.Web and AdminApp.Databases


.github/workflows/build.yml
```diff
@@ -3,7 +3,7 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-name: Build and Test
+name: Build, Test and Pack
 
 on:
   push:
```
semalaiappan, line 9: This is based on pull request based on main . Why do you want to want to pack and publish the packages . Just build and test is not enough in this build.yml

semalaiappan (changes requested):

.github/workflows/build.yml
andonyns: I will change and divide into two to keep the Build and Testing and the packaging separated, executing only on main

andonyns:
This PR will be created again once we have all the changes in place since there are more changes required. This will be divided into two tickets

PR #287, opened 2022-06-17
[AA-1468]  Adding GitHub Actions to pack and publish

andonyns:
- Adding GitHub Action to pack into NuGet package.
- Adding Action to upload to Azure Artifacts.
- Refactor build script to separate build and publish actions to improve performance.


.github/workflows/promote-release.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: pwsh
+
+    steps:
+    - name: Promote Packages
+      run: |
+        $Packages =
```
pmcvtm, line 37: MUST: Maybe I'm missing something, but is this line incomplete? Or maybe can be removed given line 46?

pmcvtm (commented):
This looks good, but I _think_ we have a syntax error that needs fixing before this will work.

pmcvtm (approved):
Thanks!

PR #293, opened 2022-06-28
[WIP] [AA-1543] Implement versioning into Admin API

CSR2017:
@pmcvtm @saa14 
Please help review the changes.
1. Changes implemented to enable the endpoint to work with one api version.
2. MapApiVersion extension added to enable version specific grouping and update the endpoint url with version (ex: /v1/vendors)

Updated more implementation details on the ticket.

Thanks


pmcvtm:
Thanks for the foundational work here y'all. I opened a new PR (#311) since while revealing, much of this ended up not being a great fit for a solution as we were looking for it to function. This will be good to have for reference as the ASP.NET versioning library evolves with minimal APIs.

PR #298, opened 2022-07-08
[AA-1469] - Docker Deployment For AdminApi

saa14:
**Description**
- Added 'Authority' authentication appsetting. This is to allow different configurations for IssuerUrl and Authority appsettings for authentication.
- Added Docker configuration files and script for Admin Api container against SQL Server.
- Added Docker configuration files and script for Admin Api container against Postgres.
- Added generate-certificate.sh to create self-signed SSL certificates which can be used when testing the containers.
- Added docker compose file and corresponding env.example for Postgres flow.
- Added AdminApi relatted container functions for testing in build.ps1. Renamed AdminApp related container functions to avoid confusion. Added BuildAdminApiDockerDevelopment.ps1 to facilitate quick dev testing inside AdminApi docker container.

### DIFF
--- a/Application/EdFi.Ods.Admin.Api/Docker/mssql/appsettings.template.json
+++ b/Application/EdFi.Ods.Admin.Api/Docker/mssql/appsettings.template.json
@@ -1,0 +1,39 @@
+{
+  "AppSettings": {
+      "DatabaseEngine": "SqlServer",
+      "DefaultOdsInstance": "EdFi ODS",
+      "ProductionApiUrl": "$API_INTERNAL_URL",
+      "ApiExternalUrl": "$API_EXTERNAL_URL",
+      "SecurityMetadataCacheTimeoutMinutes": "10",
+      "ApiStartupType": "$API_MODE",
+      "LocalEducationAgencyTypeValue": "Local Education Agency",
+      "PostSecondaryInstitutionTypeValue": "Post Secondary Institution",
+      "SchoolTypeValue": "School",
+      "PathBase": "$ADMINAPP_VIRTUAL_NAME",
+      "GoogleAnalyticsMeasurementId": "",
+      "ProductRegistrationUrl": "https://edfi-tools-analytics.azurewebsites.net/data/v1/"
+  },
+  "Authentication": {
+      "Authority": "$AUTHORITY",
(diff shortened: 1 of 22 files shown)


Application/EdFi.Ods.Admin.Api/Docker/mssql/appsettings.template.json
```diff
@@ -0,0 +1,39 @@
...
+      "DatabaseEngine": "SqlServer",
+      "DefaultOdsInstance": "EdFi ODS",
+      "ProductionApiUrl": "$API_INTERNAL_URL",
+      "ApiExternalUrl": "$API_EXTERNAL_URL",
+      "SecurityMetadataCacheTimeoutMinutes": "10",
+      "ApiStartupType": "$API_MODE",
+      "LocalEducationAgencyTypeValue": "Local Education Agency",
+      "PostSecondaryInstitutionTypeValue": "Post Secondary Institution",
+      "SchoolTypeValue": "School",
+      "PathBase": "$ADMINAPP_VIRTUAL_NAME",
```
pmcvtm, line 12: It doesn't look like this is used anywhere - in Admin App it is used to configure the app in Startup. We should either make use of it or remove it. Seems like this would tie in with the reverse proxying as well. Unclear if it relates at all to the ssl issues.

pmcvtm:
Added a commit to utilize the "PathBase" the same as done in Admin App - it appears to apply correctly in swagger but will follow up on it to verify. Additionally ignored some of the new local-only files.

pmcvtm (approved):
Approving this to move along - I am having a permissions error building the image but since it's working for @saa14, @andonyns, and an upcoming GH Action I am assuming it's an issue with my setup.

We have a known issue with SSL and will be resolving that, as well as updates to make this more closely match the patterns in [Ed-Fi-ODS-Docker](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Docker)[Ed-Fi-ODS-Docker](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Docker).

pmcvtm:
E2E Test failure is related to ODS docker - not to these changes, so we will go ahead and merge.

PR #300, opened 2022-07-16
[AA-1546] Admin API Root Page / Health Check

andonyns:
- Adding health endpoint.
- Adding Docker configuration for health check.
- Returning JSON payload in landing page.
- Pretty printing JSON result.
- Adding postman test for root endpoint.


pmcvtm (commented):
Looks good, and nice catch on the `APP`/`API` rename in Docker. I left a couple comments for cleanup - some deeper code would-be-nices that I will tack on to the end, but this all looks and works great.

pmcvtm (approved):
Added those last two commits. My Docker has been crashing so will need separate QA for the health checks, but it's working from an endpoint perspective.

PR #302, opened 2022-07-26
[AA-1552] - Fix SSL certificate issue for Docker deployment

saa14:
**Description**
- Refactored Docker files to remove the redundant '/AdminApi' nesting on the containers and rearranged the docker instructions. Refactored run.sh scripts to remove '/AdminApi' nesting. Removed env variables related to SSL certifcates.
- Removed UseHttpsRedirection to be more in line with AdminApp. Added UseForwadedHeaders to apply forwaded headers to the current request.
- Added nginx Dockerfile and configuration for AdminApi.
- Updated Docker compose file with nginx configuration. Updated SwaggerUI options to allow both Pathbase and versioning to work.
- Upgraded Swashbuckle to ensure Pathbase settings work correctly. Downgraded Microsoft.AspNetCore.Authenticaton.JwtBearer package to avoid 'method not found' exception with Docker deployment.
- Added "EnabledockerEnvironment" appsetting to disable Https requirements for docker communication between containers. Disabled TLS requirement for OpenIddict server and turned off RequireHttpsMetaData to allow communication between nginx container and adminapi container.


Application/EdFi.Ods.Admin.Api/Infrastructure/Security/SecurityExtensions.cs
```diff
@@ -46,7 +49,13 @@ public static class SecurityExtensions
                 }
 
                 opt.RegisterScopes(SecurityConstants.Scopes.AdminApiFullAccess);
-                opt.UseAspNetCore().EnableTokenEndpointPassthrough();
+                if (isDockerEnvironment)
```
pmcvtm, line 52: COULD: Not important, but instead of repeating the AspNetCore initialization, we _could_ extract the variable from the shared and then wrap just `aspNetCoreBuilder.DisableTransportSecurityRequirement()` in that condition.

Application/EdFi.Ods.Admin.Api/Infrastructure/Security/SecurityExtensions.cs
```diff
@@ -72,6 +81,11 @@ public static class SecurityExtensions
                 ValidIssuer = issuer,
                 IssuerSigningKey = signingKey
             };
+            if (isDockerEnvironment)
+            {
+                opt.RequireHttpsMetadata = false;
```
pmcvtm, line 86: COULD: Another nitpick, but instead of wrapping this in a condition we could say `opt.RequireHttpsMetadata = !isDockerEnvironment`

pmcvtm, line 86: QUESTION: Now that we're not using HttpsRedirection, does this cause an issue when not using Docker but also not using HTTPS?

If it does, is it have a helpful message?

pmcvtm (commented):
These changes look sound, I left a couple nitpicks and a question about support /alignment for HTTPS but will leave testing and approval to @CSR2017.

I want to call out that it _looks_ scary to not require HTTPS, but this puts us in step with [ODS Docker](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-Docker). Folks who want TLS on their intra-container traffic will adapt from this starting point.

CSR2017 (approved):

PR #303, opened 2022-07-28
[AA-1555] Allow to promote individual packages

andonyns:
- Fixes issue that only allows to promote all packages at the same time.
- Update postman collection to remove references to old unused variable that makes the test fail when default user is not already present.

### DIFF
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -15,6 +15,17 @@ on:
         - Prerelease
         - Release
         required: true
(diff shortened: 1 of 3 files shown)


pmcvtm (approved):
Looks great! Having that selection for the workflow is very cool, good leveraging of the features here that seems so streamlined compared to "back in the day" where that would mean copy/pasting package names or creating separate builds.

Also nice catch on the Postman change - pulled them down and everything is ✔️

PR #304, opened 2022-07-28
[AA-1544] Swagger Docs Cleanup

pmcvtm:
Fixes up and streamlines Swagger Documentation using combination of library features + expanding on custom builder from #301.

I did _not_ pull the swagger setup out to a separate extension method in an effort to make for an easier diff, but I would like to do so in a follow-up PR.

Documentation Changes
- adds success response bodies + response schemas
- fixes conventional response codes per http verb
- cleans up names + descriptions

Code Changes
- breaks all-encapsulating "default" into builder methods
  - some defaults are preserved
- replaces manual group "tagging" with conventional route-based tags
- replaces custom filters and attributes with `Swashbuckle.Annotations` and/or built-in ASP.NET methods
- replaces global response code filter with fluent syntax method
- removes no-op filters
- inline single-use constants
- inline route constants to avoid mixed bracket use in URLs
  - `"${interpolationValue}/"+"{routeValue}/"` becomes `"literal/{routeValue}/"`
- switches "required" setting out for "optional" given "required" is the default


Application/EdFi.Ods.Admin.Api/Features/Connect/ConnectController.cs
```diff
@@ -3,17 +3,19 @@
...
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using OpenIddict.Server.AspNetCore;
+using Swashbuckle.AspNetCore.Annotations;
 
 namespace EdFi.Ods.Admin.Api.Features.Connect;
 
 [AllowAnonymous]
+[SwaggerResponse(400, "Bad Request. The request was invalid and cannot be completed. See the response body for details.")]
```
saa14, line 17: We can use the `FeatureConstants.BadRequestResponseDescription` here as well.

pmcvtm: Good catch, done!

pmcvtm:
It looks like we "broke" the ordering of the endpoints in Swagger when we added the versions (the GETs are no longer grouped together). I am taking a look at that now!

saa14 (approved):

PR #314, opened 2022-08-16
[AA-1553] Update postgresql-client version

CSR2017:
@pmcvtm 
Thanks


pmcvtm (approved):
Good fix! Glad this happened right before release instead of some time later.

pmcvtm:
Just noting the ODS composition is failing to run so the one action is failing. Will look into it, may be the same issue.

PR #316, opened 2022-08-22
[AA-1510] Claimset Relationship Display

pmcvtm:
Restores the "branch" `L`-looking icon which indicates child relationships when editing a claimset, previously missing when a child resource was added to an existing parent resource that had no children at load-time.

**The Fix**

Rather than try to re-organize newly added resources from the main dropdown to be under the appropriate category, I made a change to ensure the expand-arrow and "child" dropdown is present for _all_ parent resources (not just resources which have children saved on them already). This was straightforward in code since the query is split between parent and child resources already.


saa14 (approved):

saa14:
The changes look great! Good thinking on the fix to avoid re-organizing newly added resources.

PR #319, opened 2022-08-29
[AA-1513] - Update AppCommon on AdminApp Installer

saa14:
**Description**
- Update AppCommon version on build-package.ps1 to 3.0.0.
- Added the RequiredDotNetHostingBundleVersion variable with the value 6.0.0 set and passed it as a parameter for the updated AppCommon's Install-EdFiApplicationIntoIIS call.
- Switched on the IsCustomLogin flag on Add-SqlLogins calls.


pmcvtm (approved):
Looks great!

PR #320, opened 2022-08-29
[AA-1478] - Vendor NamespacePrefix Empty String fix

saa14:
**Description**
- Added an "empty string" check on AddVendorCommand and EditVendorCommand to avoid trying to add empty strings as namespace prefixes. These strings will simply be avoided.
- Added a call to "Trim" on EditVendorCommand to remove leading and trailing spaces and be in parity with AddVendorCommand.
- Added AddVendor and EditVendor unit tests to cover the "empty" namespace prefix case.


pmcvtm:
MUST: Looks like this is fixed for "empty" but still fails when whitespace is entered. We either need to `Select` and `Trim()` each item _before_ we check `IsNullOrEmpty()` or use `IsNullOrWhitespace` instead.

pmcvtm (approved):
I made changes to account for the above comment and added test cases for that as well as null and empty entries, as well as a refactoring. Will hold off on merging till you've had a chance to check it out.

saa14:
> I made changes to account for the above comment and added test cases for that as well as null and empty entries, as well as a refactoring. Will hold off on merging till you've had a chance to check it out.

The changes look good. Thanks for pointing out the missed case and adding the needed changes.

PR #322, opened 2022-09-01
[AA-1513] Fix Installer Versions

pmcvtm:
Fixes installer version issue discovered while testing packaged AA-1513 on fresh environment.

- updates Admin App package version to most recent
  - will need to be updated again before release
- reduces `global.json` version to match version checked with AppCommon


pmcvtm:
Not _directly_ related to this fix, but want to note this issue has inspired a dive into our publish process here and a future spike to tightly couple the Admin App & Installer versions so they stay in sync without manual updates.

CSR2017 (approved):

PR #324, opened 2022-09-07
Update command in order to select the firsts element in the screen

jpardogrowthaccelerationpartners:
Update search selectors to take the first element they find in the region. Now it happens not every time that it finds two selectors with the same name and the test fails.

This change should not affect running in a clean environment or one that already has data.

Screen Shot 2022-09-07 at 14 37 35


andonyns (approved):
Good idea

PR #328, opened 2022-09-08
[AA-1556] Disable Product Improvement By Configuration

pmcvtm:
Introduces a new app settings flag which disables the "Product Improvement" configuration page and sets it to false.

- app settings flag defaults to `true`
- when flag is set to `false`:
  - configuration always returns product improvement is disabled
    - this handles scenario in which config was previously set to true,  before changing the flag
  - bypasses first-time-setup config screen and sets configuration to false during initialization
  - removes "⚙Configuration" link from layout and blocks direct access to page


saa14 (approved):
The changes and comprehensive tests look great.

PR #329, opened 2022-09-08
[AA-1374] - Validate ODS/API before trying to communicate with it

saa14:
**Description**
- Added OdsApiValidator.cs to act as the validator service for ODS API and perform a JSON schema check for the configured ODS API Url.Added NJsonSchema package for JSON schema validation.
- Used IOdsApiValidator to validate the API before trying to communicate with the API using a client.
- Added unit tests for OdsApiValidator.cs.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Management.Tests/Controllers/EducationOrganizationsControllerTests.cs
+++ b/Application/EdFi.Ods.AdminApp.Management.Tests/Controllers/EducationOrganizationsControllerTests.cs
@@ -34,6 +34,7 @@ namespace EdFi.Ods.AdminApp.Management.Tests.Controllers
         private Mock<ITabDisplayService> _tabDisplayService;
         private IInferExtensionDetails _inferExtensionDetails;
         private Mock<IInferExtensionDetails> _mockInferExtensionDetails;
+        private Mock<IOdsApiValidator> _mockOdsApiValidator;
 
         [SetUp]
         public void Init()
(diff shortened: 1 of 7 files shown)


pmcvtm:
MUST: We need to be more granular with our error evaluation and messaging. Often the URL is configured correctly but something else wrong that we can discern and return a specific message to help the user. It's good to have this level of schema evaluation, but this check happens before other steps in the ODS/API communication chain, we at minimum need to support the same error handling as those steps, to not lose them. See the handling in `TokenRetriever`, for example.

We can determine some helpful messages to return to the user based on what responses come back from the request, or what exceptions get thrown, either in `SimpleGetRequest` or in the validator itself. Here are a few examples:

- ODS not running would throw `HttpRequestException` with a specific message
- ODS service failing would return 503 response (this surfaces as "unable to parse" currently)
  - Other failing response codes might at least indicate what the response was in the message
- Missing or invalid SSL would throw `HttpRequestException` with a specific message
- Malformed URIs would throw a `InvalidOperationException` with a specific message
- Unexpected errors should return the `Exception.Message` in the subsequent `throw` so they can troubleshoot or report effectively (iow, don't swallow errors entirely)

pmcvtm (approved):
These updates look excellent! Thank you for expanding the handling there.

pmcvtm:
Looks like we have a setup failure for E2E tests. I am going to merge this and we'll address that separately.

PR #330, opened 2022-09-09
[AA-1579] Fix E2E Test Setup for ODS 6

pmcvtm:
Removes previous workaround needed before ODS/API upgraded to .NET 6. Now that that change is in place, this workaround is not only unneeded but causing our setup to fail.


pmcvtm:
Additionally looks like that same step was failing due to the version change to 2.4.**37**. Existing regex was matching single-digits only. This is now fixed as well.

saa14 (approved):

PR #331, opened 2022-09-09
[AA-1574] Update Learning standard core package version to fix Polly package version conflict error

CSR2017:


pmcvtm (approved):
Looks like it's working again! Thanks for going through all the steps to get this upgrade.

PR #332, opened 2022-09-12
[ODST-214] - Fixes created school within some LEA not being displayed under that LEA

saa14:
**Description**
Downgraded AutoMapper (from 11.0.1 to 10.1.1) and AutoMapper.Extensions.Microsoft.DependencyInjection (from 11.0.0 to 8.1.1) packages as a temporary fix to the issue with LocalEducationAgencyReferenceResolver not being called during mapping.


pmcvtm (approved):
Looks like this is the fix! Just noting here that @saa14 created a follow-up ticket to further diagnose the issue and re-upgrade.

PR #333, opened 2022-09-13
[ODST-216] - Upgrade AutoMapper and Fix for LocalEducationAgencyReferenceResolver issue

saa14:
**NOTE: This is related to the temporary fix introduced in [ODST-214](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/pull/332) and so warrants a test to the issue highlighted in that ticket as well.**
**Description**
- Upgraded AutoMapper (to the latest 11.0.1) and AutoMapper.Extensions.Microsoft.DependencyInjection (to the latest 11.0.0) packages.
- Added static Resolve() overloads for LocalEducationAgencyReferenceResolver and StateEducationAgencyReferenceResolver.
- Added missing "CtorParam" mapping configuration to fix the Resolver not being called issues.


pmcvtm (approved):
Looks good and still works, thanks for the quick turnaround on the 'real' fix here!

PR #334, opened 2022-09-14
SPIKE DRAFT: provenance and SBOM generation

stephenfuqua:
This is not for merging - only for review and discussion. For context, please see [Supply Chain Integrity](https://techdocs.ed-fi.org/display/EDFIODS/Supply+Chain+Integrity) in Tech Docs.

Most of the Actions yml files are renamed to disable them during my testing. I tested on my fork, where I could merge to main and create releases as often as I wanted.

End result: see this fake release, [v1.2.5](https://github.com/stephenfuqua/Ed-Fi-ODS-AdminApp/releases/tag/v1.2.5), which has SBOM manifest and provenance jsonl files attached to the release. In the [actions tab](https://github.com/stephenfuqua/Ed-Fi-ODS-AdminApp/actions), you can see that this latest run was generated from a tag rather than a branch. The tag was in turn created from a new GitHub release. So that's the workflow: create a release, and the provenance and SBOM files will be attached to the release. And [here's the action log](https://github.com/stephenfuqua/Ed-Fi-ODS-AdminApp/actions/runs/3053577102) for the release.

image


.github/workflows/publish.yml
```diff
@@ -52,44 +79,137 @@ jobs:
...
-          }
-          $arguments.PackageFile = $artifact
-          ./build.ps1 Push @arguments
+          # sha256sum returns "<hashcode>  <name of file". Split that and return only the <hashcode>.
+          sbom_hash=$(sha256sum ./manifest/${{ env.MANIFEST_FILE }} | awk '{split($0,a); print a[1]}')
+          echo "::set-output name=sbom-hash::$sbom_hash"
+
+  sbom-release:
+    runs-on: ubuntu-latest
+    needs: [sbom-create]
```
andonyns, line 136: This can go without the brackets as the rest of jobs

.github/workflows/publish.yml
```diff
@@ -52,44 +79,137 @@ jobs:
...
+
+  provenance:
+    needs: pack
+    permissions:
+      actions: read
+      id-token: write
+      contents: write
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    # This is failing :-(
+    # uses: ed-fi-alliance-sso/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
```
andonyns, line 163: The path here should be ed-fi-alliance-oss/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0

.github/workflows/publish.yml
```diff
@@ -52,44 +79,137 @@ jobs:
...
     runs-on: ubuntu-latest
-
-    defaults:
-      run:
-        shell: pwsh
-
+    needs: pack
+    permissions:
+      actions: read
+      id-token: write
```
andonyns, line 87: Is this permission required? I tested this on my branch without the permission and the workflow finished correctly. Wondering for this one and the pack and provenance jobs

.github/workflows/publish.yml
```diff
@@ -17,32 +19,57 @@ env:
...
+  MANIFEST_FILE: "_manifest/spdx_2.2/manifest.spdx.json"
 
 jobs:
   pack:
-
     runs-on: ubuntu-latest
-
     defaults:
       run:
         shell: pwsh
```
andonyns, line 30: Now that there are a few scripts on bash, I wonder if it is better to default to bash (the original) and specify PowerShell on the required scripts.

.github/workflows/publish.yml
andonyns: I tested this on my fork and it worked correctly

stephenfuqua: Oh, I see that I wrote "sso" instead of "oss" 😆

.github/workflows/publish.yml
stephenfuqua: I didn't test this carefully after copying some permissions from the SLSA action. Thinking about it, this definitely should _not_ be needed.

.github/workflows/publish.yml
```diff
@@ -17,32 +19,57 @@ env:
...
+    outputs:
+      hashes: ${{ steps.hashes.outputs.hashes }}
 
     steps:
     - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
+
```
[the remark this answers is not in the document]
andonyns: I really like that, for the future, it might be good to add a linting verification/auto-fix for yml files :)

PR #337, opened 2022-09-22
[AA-1703] - Add SSO support to Admin App

saa14:
**Description**
**Web and Management Projects**
- Added OpenIdConnect and Cookies authentication packages.
- Added IdentitySettings section to appsettings.json. Added corresponding classes/constants for IdentitySettings.
- Added OpenIdConnect specific OpenIdConnectUserMustExistHandler for the user authorization requirement.
- Added OpenIdConnectController and corresponding login/logout related actions
- Refactored Startup.cs configuration for EntityFramework Identity and brought it out to a separate function. Added configuration for OpenIdConnect.
- Added AddUserLoginModel and AddUserLoginCommand to facilitate adding the user and userlogin records into the database on OIDC login. Added OpenIdConnectLoginService to facilitate setting up the user in the Identity tables and assign SuperAdmin role on an OIDC login.
- Added an OpenIdConnect specific UserContextFilter to populate the user roles and permissions based on the oidc information available. Separated out the registration for UserContextFilter to only Identity flow. Registered OpenIdConnectUserContextFilter for the OIDC flow.
- Added conditional logic to disable Users tab for OIDC flow.Refactored to use UserId from the usercontext to ensure instance registration works with OIDC login flow.

**Tests Project**
- Added IdentitySettings section to the Management.Tests/appsettings.json.Added tests for AddUserLoginCommand and OpenIdConnectLoginService.


Application/EdFi.Ods.AdminApp.Management/Helpers/IdentitySettings.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    public class OpenIdSettings
+    {
+        public string Authority { get; set; }
+        public string ClientId { get; set; }
+        public string ClientSecret { get; set; }
+        public string UserProfileUri { get; set; }
+        public string AuthenticationScheme { get; set; }
```
pmcvtm, line 26: COULD: Would it make sense (and work) to rename this to `OidcProvider`? That may bypass some of the issues with changing providers out. We _know_ it's OIDC if they're using these options.

For context, I swapped which KeyCloak instance I am using and now can't sign in with the same email.

pmcvtm: We may need to also include that in other queries.

Application/EdFi.Ods.AdminApp.Web/Startup.cs
```diff
@@ -169,6 +167,114 @@ public void ConfigureServices(IServiceCollection services)
...
+                    context.HttpContext.RequestServices.GetRequiredService<IOpenIdConnectLoginService>();
+
+                var claimsIdentity = context.Principal?.Identity as ClaimsIdentity;
+                var oidcUserId = claimsIdentity?.Claims.FirstOrDefault(m => m.Type == ClaimTypes.NameIdentifier)?.Value;
+                var oidcUserEmail = claimsIdentity?.Claims.FirstOrDefault(m => m.Type == ClaimTypes.Email)?.Value;
+
+                if (openIdConnectLoginService != null && oidcUserId != null && oidcUserEmail != null)
+                {
+                    var oidcAuthScheme = identitySettings.OpenIdSettings.AuthenticationScheme;
+                    var identityUserId = await openIdConnectLoginService.AddUserLoginForOpenIdConnect(oidcUserId, oidcUserEmail, oidcAuthScheme, oidcAuthScheme);
```
pmcvtm, line 271: SHOULD: Since this call happens on every sign-in, we might either:

- check that the user login exists and bypass (possibly setting relevant values from the identity user)
- or rename the method to imply it gets run every time

Application/EdFi.Ods.AdminApp.Web/Helpers/OpenIdConnectLoginService.cs
```diff
@@ -0,0 +1,67 @@
...
+            var addUserLoginModel = new AddUserLoginModel
+            {
+                UserEmail = oidcUserEmail,
+                LoginProvider = loginProvider,
+                ProviderDisplayName = providerDisplayName,
+                ProviderKey = oidcUserId
+            };
+
+            var userLoginModelValidator = new AddUserLoginModelValidator(_identity);
+            var userLoginModelIsValid = (await userLoginModelValidator.ValidateAsync(addUserLoginModel)).IsValid;
```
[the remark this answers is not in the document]
pmcvtm: As a side-note, if we elect to not check for an existing `UserLogin` before running this service in the outside context, we should pull that validation rule out, so it does _not_ get logged extraneously.

Application/EdFi.Ods.AdminApp.Web/OpenIdConnectUserMustExistHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+                var userLogin = _identity.UserLogins.SingleOrDefault(
+                    x => x.LoginProvider == _identitySettings.OpenIdSettings.AuthenticationScheme &&
+                         x.ProviderKey == userId);
+
+                if (userLogin != null)
+                {
+                    context.Succeed(requirement);
+                }
+                else
+                {
```
pmcvtm, line 43: MUST: I believe we need to call `context.Fail()` here to ensure the user does _not_ sign in. Currently it gets into a weird semi-signed in state as the process continues.

pmcvtm: To be safe we should also `try` around _all_ of the handling and `Fail()` and `throw;` in the `catch`

Application/EdFi.Ods.AdminApp.Web/OpenIdConnectUserMustExistHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                var userLogin = _identity.UserLogins.SingleOrDefault(
+                    x => x.LoginProvider == _identitySettings.OpenIdSettings.AuthenticationScheme &&
+                         x.ProviderKey == userId);
+
+                if (userLogin != null)
+                {
+                    context.Succeed(requirement);
+                }
+                else
```
pmcvtm, line 42: SHOULD: When a user does not have an email address set in their OIDC provider, this requirement will fail at this point. That's OK, but it would be nicer to check early and instead let the user know that it's not their fault: "To use Admin App, users must have an email address set in their login provider system. Contact your administrator to resolve this issue. System error: `{exception message}`" with the actual exception set as the `inner`, too.

pmcvtm (commented):
Overall this is working well but we have a few places to tighten up which I think will make this easier for users to set up and catch some unexpected cases.

There's an enhancement we should make before merging as well, but I will lay that out as a separate comment.

pmcvtm (approved):
Thanks for making those edits! I am about to push up some other changes.

The big one is **requiring a role claim**, as we discussed with the team. This uses constant values added to the `Role` rich enum  to determine and save the roles to the claim (and in the DB). Given the approach, it didn't take much effort to map to the literal roles and support both Super Admin and regular Admin from the get-go, so that's nice.

I also made some other fixes:
- Fixing Identity User creation to use the _configured_ claim names rather than the ones we set
  - Added separation between [display] name, email, and identifier claim types. If someone wants to set all those to "email," they can
- Allowing multiple logins for the same user
  - This is something that we said we said we might do later but it made my testing annoying without, so went ahead and changed it.
  - This also prevents some extra validation warning logging which weren't _really_ issues
- Moved where the "are claims set correctly check" to happen on sign-in
  - The `Fail` in the auth handler didn't change much, that was my bad. Better to fail loudly when we try to create the use.r

I'm going to ✔️ Approve for now, but will wait for you to merge @saa14, let me know if you have any concerns with the new commits.

pmcvtm:
Another update: @saa14 discovered the `throws` happening in the `OnSignIn` event handler are not picked up by the global error handling, so the setup issues were not apparent to end users. I've adjusted the handling there so:

- invalid config or claims do not create a user
- the `UserMustExist` handler returns nice messages again
- role claims are required by a default `ClaimPolicy` for OIDC setup
- default `AuthorizationFilter` is applied globally to prevent invalid users from getting around

I also figured out how to fix the logout! @saa14 can you give another check now?

saa14:
> Another update: @saa14 discovered the `throws` happening in the `OnSignIn` event handler are not picked up by the global error handling, so the setup issues were not apparent to end users. I've adjusted the handling there so:
> 
> * invalid config or claims do not create a user
> * the `UserMustExist` handler returns nice messages again
> * role claims are required by a default `ClaimPolicy` for OIDC setup
> * default `AuthorizationFilter` is applied globally to prevent invalid users from getting around
> 
> I also figured out how to fix the logout! @saa14 can you give another check now?

Thanks for the updates and fixes related to **role claims** and **sign out**. I have tested out the changes with Google as the login provider (with a hardcoded role claim) and it seems to work as expected. I am still facing a probable configuration-related issue on the logout and will further investigate that. In the meanwhile, this PR seems good to merge.

PR #346, opened 2022-10-06
[AA-1600] OIDC AppSettings Truncation + Other Issues

pmcvtm:
Fixup of issues found during testing of #337 

- Fixes mangled `appsettings.json` caused by truncation during build
- Adds missing action for friendly `Unauthorized` page
- Fixes claim -> cookie mapping during sign-in
  - Claims were being used to set user info in DB correctly, but not persisting to Principal


CSR2017 (approved):
Changes look great. Thank you

PR #349, opened 2022-10-06
[AA-1591] Rebuild Actions Workflows

stephenfuqua:
* Remove TeamCity
* Build SBOM files
* Build Provenance files
* MinVer for version numbers
* New chain of workflows and jobs to support a more automated release process


pmcvtm (approved):
This looks great. I think we all expect some fine tuning and enhancements / expansions in the coming months but this looks great for getting the SBOM and Provenance in as well as getting a jump on release number control!

Noting here also we plan on using this for 3.0.0 for Web/DB but release and for now run with the existing `Pack and Publish Installer` workflow for the Installer.

andonyns (approved):

PR #350, opened 2022-10-07
[AA-1591-3] Prevent running on alpha

stephenfuqua:


.github/workflows/on-merge-or-tag.yml
```diff
@@ -46,6 +48,7 @@ jobs:
           "::set-output name=admin-api::$($apiPRefix)$(minver -t $apiPRefix)" | Write-Output
 
       - name: Create Admin App Web Pre-Release
+        if: "! startsWith(github.ref_name, 'Admin.Api-v')"
         run: |
           $version = "${{ steps.versions.outputs.admin-app }}"
```
stephenfuqua, line 53: Let's say we want to release Admin.Api: push a tag called `Admin.Api-vX.Y.Z`. That will trigger this workflow. Ultimately we want it to create a new pre-release (which we'll manually flip to release, at least for now) - but only of Admin API! This condition on line 51 prevents the Admin.Api tag from creating a new pre-release of AdminApp, which would have ended up being a duplicate of an existing one.

.github/workflows/on-prerelease.yml
```diff
@@ -90,7 +90,7 @@ jobs:
           $apiVersion = "${{ steps.versions.outputs.admin-api-semver }}"
 
           "Building ${{ steps.package-name.outputs.package-name }} NuGet Package" | Write-Output
-          if ("${{ steps.package-name.outputs.package-name }}" -eq "AdminApp") {
+          if ("${{ steps.package-name.outputs.package-name }}" -eq "AdminApp.Web") {
```
stephenfuqua, line 93: Dumb mistake that I failed to catch before because I wasn't paying enough attention to the NuGet packages in my prior testing.

.github/workflows/on-release.yml
```diff
@@ -51,8 +51,6 @@ jobs:
                 if ($_.tag_name -like "$($packageName)-*" -and $_.prerelease) {
                     "Deleting pre-release $($_.tag_name)" | Write-Output
                     Invoke-RestMethod -Method Delete -Uri $_.url -Headers $gh_headers
-
-                    &git tag -d $_.tag_name
```
stephenfuqua, line 55: Deleting the release is successfully deleting the tag now. I had thought I saw the tag being left behind at one point, but it is fine now.

Perhaps it works now because I had to give the EdFiBuildAgent `maintain` level access, in order to allow the build agent to push tags.

.github/workflows/on-release.yml
```diff
@@ -63,6 +61,9 @@ jobs:
     name: Promote Azure Artifact
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout the Actions repo
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846  # v3.0.0
+
```
stephenfuqua, line 66: The next step uses one of the `.eng/*.ps1` scripts, therefore the repo must be checked out.

.github/workflows/on-prerelease.yml
```diff
@@ -264,7 +264,6 @@ jobs:
         run: iex "& { $(irm https://aka.ms/install-artifacts-credprovider.ps1) } -AddNetfx"
 
       - name: Push Package to Azure Artifacts
-        if: "! github.event.pull_request.head.repo.fork"
```
stephenfuqua, line 267: This condition did not work - the step still ran in my fork. However, I don't have an Azure Artifacts key setup in my fork, so the step failed and I'm fine with that.

andonyns: Good to know that it does not work

andonyns (approved):

pmcvtm (approved):

PR #351, opened 2022-10-07
[AA-1526] Security Scans on Pull Request

stephenfuqua:
* Dependency review (C# and Actions)
* CodeQL
* BIDI Trojan Attack detection


andonyns:
Adding the approved list action is making the build fail because they are some actions not approved. I will update dependencies in this PR to make the build pass

.github/workflows/on-pullrequest-C#.yml
```diff
@@ -3,25 +3,33 @@
...
     paths:
-      # TODO: restore this with AA-1601
+      # TODO: restore this with AA-1601, except run in a PS-specific
+      # workflow, so that the C# build and CodeQL do not run
+      # unnecessarily.
       # - "**/*.ps1"
       # - "**/*.psm1"
       - "**/*.cs"
       - "**/*.csproj"
+      - ".github/**/*.yml"
```
stephenfuqua, line 20: @andonyns I noticed that the E2E tests are running for any pull requests. What do you think about adding path limits like the above, so that they only run for C# changes? Or alternately, could trigger "after the on pull request - C#" job runs.

Oh, one other change will be required - the existing "after pull request yml" needs to be updated to have this new workflow name (with " - C#" appended).

andonyns: I agree, I will include this into this PR

andonyns:
This should be merged after https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-Actions/pull/28 since the PR includes new approved actions

.github/workflows/on-pullrequest-C#.yml
```diff
@@ -3,25 +3,33 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-name: On Pull Request
+name: On Pull Request - C#
```
andonyns, line 6: Although nothing wrong with that, I'm not totally on board of renaming the files per language, I can be easily convinced though 😄 
I wonder if there's a way that we can have one "On Pull Request" workflow and have specific jobs by format of files changed, and run only that job, instead of having multiple On Pull Request - Language. (Which I think, worst case scenario could be 2 or 3)

.github/workflows/on-pullrequest-C#.yml
```diff
@@ -3,25 +3,33 @@
...
 jobs:
   # TODO: restore this with AA-1601
   # run-ps-lint:
   #   name: PowerShell Linter
   #   uses: Ed-Fi-Alliance-OSS/Ed-Fi-Actions/.github/workflows/powershell-analyzer.yml@main
+  scan-actions-bidi:
+    name: Scan Actions, scan all files for BIDI Trojan Attacks
+    uses: ed-fi-alliance-oss/ed-fi-actions/.github/workflows/repository-scanner.yml@main
+    with:
+      config-file-path: ./.github/workflows/bidi-scanner/config.json
```
andonyns, line 32: This is [failing](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/actions/runs/3206166510/jobs/5239575875) because the config.json file does not exist 

Is this config file required?

stephenfuqua: No, and I need to drop that. Was specific to Analytics Middle Tier.

.github/workflows/on-pullrequest-C#.yml
stephenfuqua: https://stackoverflow.com/questions/58808690/conditional-jobs-and-steps-based-on-updated-paths

This suggests a hack using `git diff`. Can set an output variable and use that to trigger a job. I agree would prefer not to have separate "on-pullrequest" files. So I'll rename this and we can try that hack approach when we do the psscript analyzer.

pmcvtm (approved):
Looks good! Important to have checks like this. Also nice to get the whitespace consistent, though it made for a pretty misleading diff about the scope of these changes!

PR #352, opened 2022-10-07
[AA-1591] Need to run in pwsh

stephenfuqua:


pmcvtm (approved):

pmcvtm:
The E2E tests are failing to set up since our new appsettings are missing from the ODS-Docker repo. It's unrelated to this change and we'll update them with the upgrade to Admin App 3.0 in that repo.

PR #353, opened 2022-10-07
[AA-1585] Update Admin App package version on installer and update installer semantic version

CSR2017:


pmcvtm (approved):
Approving - a note that this is suffering from the same setup failure for E2E tests which should be resolved in the ODS Docker repo soon.

PR #355, opened 2022-10-13
[ADMINAPI-34] - Adding e2e tests for claimsets endpoint

saa14:
**Description**

**Code Fixes and Refactoring**
- Refactored to only show a single validation message for each distinct duplicate resource in the ResourceValidator.
- Refactored to throw the NotFoundException instead of validation failure exception to be similar to other not found cases for the endpoint.
- Refactored the EditClaimSet flow to throw an AdminAppException from within the claim set command and catch-rethrow a validation exception with the "System Reserved claim sets" exception message.
- Refactored the DeleteClaimSet flow to throw an AdminAppException from within the delete claim set command and catch-rethrow a validation exception with the "System Reserved claim sets" exception message.
- Added the additional condition to make sure the claimset exists when validating claimset name change.

**E2E Tests**
- Updated the AdminApi postman collection to add e2e tests for claimsets endpoint.
- Please consult the testing section on the ticket for the test cases covered.

### DIFF
--- a/Application/EdFi.Ods.Admin.Api/Docker/DB-Admin/pgsql/Dockerfile
+++ b/Application/EdFi.Ods.Admin.Api/Docker/DB-Admin/pgsql/Dockerfile
@@ -3,7 +3,7 @@
 # The Ed-Fi Alliance licenses this file to you under the Apache License, Version 2.0.
 # See the LICENSE and NOTICES files in the project root for more information.
 
-FROM edfialliance/ods-api-db-admin:2.1.4 
+FROM edfialliance/ods-api-db-admin:v2.1.4 
 LABEL maintainer="Ed-Fi Alliance, LLC and Contributors <techsupport@ed-fi.org>"
 
 ENV POSTGRES_USER=${POSTGRES_USER}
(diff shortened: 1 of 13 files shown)


pmcvtm:
Regarding the E2E failure, it looks like the Docker build is set to use a specific Admin API package rather than building an image from the current changes or doing the version substitution like Admin App does. We won't be able to get that passing in this PR without making some other changes. We can consider:

- Removing the PR check (so it only builds after a new package is released)
  - This still will only run using the hard-code version in the Dockerfile. Better option would be to utilize the [set-versions](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/blob/main/Application/EdFi.Ods.AdminApp.E2E.Tests/gh-action-setup/set-versions.sh) script
- Adding a "dev" Dockerfile which builds from the current code instead of pulling down from Nuget and utilizing that in the E2E step

Might be best to merge this regardless of that check, but I would suggest we get a passing E2E test with the package published after this change before releasing.

pmcvtm:
COULD: We might consider appending a GUID to our test claimset name. Currently if we get "halfway" through a run it can end up in an un-repeatable state. We haven't had any unique constraints in features yet so this hasn't really come up, but you can see how to create and insert GUIDs in the `Register` tests.

pmcvtm (approved):
Looks good! Great to have these missing tests and thank you for doing the fit-to-pattern fixups. I appreciate the rich validation checks with the loops, too. Nice way to get that coverage without being too brittle.

I'm approving this but will leave it unmerged in case you'd like to make the suggested change, and so the team can give an OK regarding the failing check and set a plan for release.

saa14:
> COULD: We might consider appending a GUID to our test claimset name. Currently if we get "halfway" through a run it can end up in an un-repeatable state. We haven't had any unique constraints in features yet so this hasn't really come up, but you can see how to create and insert GUIDs in the `Register` tests.

Good thought. Added the suggested GUID to claimset names.

PR #357, opened 2022-10-19
[AA-1603] - Include Installer in Admin App Web package

saa14:
**Description**
- Added `Add-AppCommon` and `Move-AppCommon` to package-manager.psm1 to download AppCommon and move it to a destination directory.
- Added `AddAppCommonPackageForInstaller` step to the BuildPackage flow in order to include pre-zipped AppCommon files to be used by the Admin App installer.
- Updated the AdminApp.Web nuspec file to include necessary script files for installation.
- Refactored `Get-AdminAppPackage` to `Set-AdminAppPackageSource`. The new function does a file version property check between the provided config version and admin source binaries used as well as setting the correct path to the binaries.
- Made the PackageSource parameter mandatory and updated the documentation for the same.
- Updated the `install.ps1`, `test.ps1` and `upgrade.ps1` scripts to include the package source parameter.
- Removed obsolete scripts which are no longer required for the installer.
- Removed the publish-installer github workflow.


pmcvtm (approved):
This looks great. We discussed offline that we'll be following up to address the follow as tech debt:

- Renaming the `EdFi.Suite3.Installer.AdminApp` folder in source (since it's not longer a separate package)
- Fixing up how we set or call `PackageVersion` (see comment for related thought)
- Cleaning up parameters and functions per new usage and obsolescence
- Refactoring tests

PR #361, opened 2022-10-31
[DI-1241] - Fix for User Unauthorized page redirect

saa14:
**NOTE: This is a fix for Admin App corresponding to [the Data Import PR](https://github.com/Ed-Fi-Closed/Ed-Fi-DataImport/pull/378)**
**Description**
Added basePath to the UserUnauthorized redirect path if it exists.

### DIFF
--- a/Application/EdFi.Ods.AdminApp.Web/Startup.cs
+++ b/Application/EdFi.Ods.AdminApp.Web/Startup.cs
@@ -362,8 +362,17 @@ namespace EdFi.Ods.AdminApp.Web
             app.UseStatusCodePages(ctx =>
             {
                 var response = ctx.HttpContext.Response;
-                if (response.StatusCode == (int) HttpStatusCode.Unauthorized || response.StatusCode == (int) HttpStatusCode.Forbidden)
-                    response.Redirect("./Home/UserUnauthorized");
+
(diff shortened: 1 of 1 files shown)


Application/EdFi.Ods.AdminApp.Web/Startup.cs
```diff
@@ -362,8 +362,17 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerF
...
+
+                if (response.StatusCode == (int)HttpStatusCode.Unauthorized || response.StatusCode == (int)HttpStatusCode.Forbidden)
+                {
+                    var basePath = response.HttpContext.Request.PathBase;
+                    var userUnauthorizedPath = "/Home/UserUnauthorized";
+                    if (basePath.HasValue)
+                    {
+                        userUnauthorizedPath = $"{basePath}/Home/UserUnauthorized";
+                    }
+                    response.Redirect(userUnauthorizedPath);
```
pmcvtm, line 374: POTENTIAL FUTURE SHOULD: Our code scanners are working! Or maybe "functional" is a better word, since this was a false-positive. It allows me to dismiss this warning as such (added a note too) but if it continues to bug us we might consider renaming this variable.

pmcvtm (approved):
This looks good! Thanks for making the change in both repositories.

Nothing here the E2E failure here is a startup issue with the *admin app* package in the workflow. Something we need to investigate but not a regression from this change, so we're disregarding here.

PR #363, opened 2022-11-16
[AA-1628] Fix Test Report Paths + Optimize

pmcvtm:
Fixes to the `OnPullRequest` workflow found during improvements spike:

- fix C# test report paths to correctly capture artifact
- fix unit tests to not rebuild
- other cleanup:
  - shorten or change names for "check" readability
  - remove comments

The test reporter fix won't work until it's merged into main, since it happens _after_ the pull request. I wonder if it is possible to determine if the PR source is from a fork, so we can take different paths for forks vs. not (not that it would help here)

### DIFF
--- a/.github/workflows/after-pullrequest.yml
+++ b/.github/workflows/after-pullrequest.yml
@@ -21,6 +21,6 @@ jobs:
         with:
           artifact: csharp-tests
           name: C# Unit Test Results
-          path: "*.trx"
+          path: "**/*.trx"
           reporter: dotnet-trx
           fail-on-error: false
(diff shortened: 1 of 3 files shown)


pmcvtm:
Part of the reason we separate our unit and e2e test reporting workflow is because it doesn't work built-in for forks. Wondering if that is also the case for these other scanners

image

saa14 (approved):

stephenfuqua:
I have no idea why those two actions won't run from a fork. For now, I suggest that @pmcvtm re-create this PR from a branch on the main repository, instead of from a fork.

Later we can investigate how to get those working from a fork. I'm not willing to turn them off right now, and have other priorities today so can't investigate immediately.

pmcvtm:
Recreated directly on the repo: #365

PR #365, opened 2022-12-02
[AA-1628] Fix Test Report Paths + Optimize (Unforked)

pmcvtm:
Fixes to the OnPullRequest workflow found during improvements spike. Recreated from #363 commits directly for this repo instead of a fork in an attempt to get checks to pass correctly.

- fix C# test report paths to correctly capture artifact
- fix unit tests to not rebuild
- other cleanup:
  - shorten or change names for "check" readability
  - remove comments

The test reporter fix won't work until it's merged into main, since it happens after the pull request. I wonder if it is possible to determine if the PR source is from a fork, so we can take different paths for forks vs. not (not that it would help here)

### DIFF
--- a/.github/workflows/after-pullrequest.yml
+++ b/.github/workflows/after-pullrequest.yml
@@ -21,6 +21,6 @@ jobs:
         with:
           artifact: csharp-tests
           name: C# Unit Test Results
-          path: "*.trx"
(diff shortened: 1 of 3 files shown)


stephenfuqua (commented):
👍

saa14 (approved):

PR #368, opened 2022-12-07
[AA-1608] Update Actions to use environment for state instead of set-output

CSR2017:
@pmcvtm @stephenfuqua 
Refactored all the set-output to $GITHUB_OUTPUT.
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
Thanks


pmcvtm (approved):
Looks great! Obviously the test will be post-merge but good to have this fixed. Also personally I find the "new way" to be clearer and more readable, so bonus!

PR #369, opened 2022-12-15
[AA-1608] GITHUB_OUTPUT fix PowerShell specific syntax

CSR2017:
@pmcvtm 
Please help review and merge the changes. Thanks


pmcvtm (approved):
Looks good! There were two other usages of `GITHUB_OUTPUT` which I appened `$env:` to. Thanks for the quick turnaround on this fix!

pmcvtm:
Reverted that commit! Missed that the syntax update is only for `pwsh` and those other usages were explicitly bash.

PR #452, opened 2024-03-21
[AA-1738] - Updates the path of missing files in Admin App E2E Test

DavidJGapCR:


jpardogrowthaccelerationpartners (approved):

.github/workflows/aa-e2e.yml
```diff
@@ -41,15 +42,15 @@ jobs:
           chmod +x ./main/Application/EdFi.Ods.AdminApp.E2E.Tests/gh-action-setup/set-versions.sh
           ./main/Application/EdFi.Ods.AdminApp.E2E.Tests/gh-action-setup/set-versions.sh
       - name: Run ODS
-        run: docker-compose -f ./ods-docker/Compose/pgsql/compose-aa-build.yml --env-file ./main/Application/EdFi.Ods.AdminApp.E2E.Tests/gh-action-setup/.automation.env up -d
+        run: docker-compose -f ./main/Application/EdFi.Ods.AdminApp.E2E.Tests/gh-action-setup/compose-aa-build.yml --env-file ./main/Application/EdFi.Ods.AdminApp.E2E.Tests/gh-action-setup/.automation.env up -d
```
andonyns, line 45: If this path is used, therefore the copy on line 37 is no longer needed

andonyns: Also, do we need NuGet in line 38? (does not affect but it might be a good moment to update)

andonyns (approved):
There are a couple of comment but those do not affect the fix, are improvement opportunities

DavidJGapCR:
Done. Thanks for your comments.

PR #483, opened 2025-02-05
[AA-1776] Update the deprecated packages

jpardogrowthaccelerationpartners:


.github/workflows/on-prerelease.yml
```diff
@@ -213,7 +213,7 @@ jobs:
       actions: read
       id-token: write
       contents: write
-    uses: Ed-Fi-Alliance-OSS/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@main
+    uses: Ed-Fi-Alliance-OSS/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v2.0.0
```
stephenfuqua, line 216: @jpardogrowthaccelerationpartners you probably  did the same thing in other pull requests and I only noticed it now: why use `main` on line 174 instead of `v2.0.0`?

jpardogrowthaccelerationpartners: Well, that was the one that was working in other repositories, that's why I left it with main. But if necessary, I can change the version.

stephenfuqua: I think that was an accident. Can you try testing in a fork to see if changing to `v2.0.0` works on line 174 as well?

stephenfuqua: But don't spend a lot of time. This is ok and we can keep it if changing causes any challenges. I might need to investigate more closely at some point in the future.

jpardogrowthaccelerationpartners: I just ran it on my fork and main and v2.0.0 fail at the same step.. so I think it won't affect if we change it to v2.0.0

Screenshot 2025-02-07 at 8 58 21 AM

stephenfuqua (approved):

PR #495, opened 2025-07-01
Potential fix for code scanning alert no. 215: DOM text reinterpreted as HTML

stephenfuqua:
Potential fix for [https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/security/code-scanning/215](https://github.com/Ed-Fi-Alliance-OSS/Ed-Fi-ODS-AdminApp/security/code-scanning/215)

To fix the issue, the `val` variable (and any other potentially tainted data) should be properly escaped before being interpolated into the HTML string. Escaping ensures that special characters like `<`, `>`, and `"` are converted into their HTML entity equivalents, preventing them from being interpreted as executable code.

The best way to fix this is to use a utility function to escape the `val` variable. For example:
1. Define an `escapeHtml` function that converts special characters into their HTML-safe equivalents.
2. Use this function to sanitize `val` before including it in the HTML string.

The changes will be made in the `CreateMultiCheckBox` function where the HTML string is constructed.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


stephenfuqua:
I'm not sure if we really want to do this. We'll have to think about the testing implications first.

jagudelo-gap (approved):

PR #513, opened 2025-07-11
[AA-1797] Multiselect button text wrapping fix

ricardogamboagap:


jleiva-gap (changes requested):
The Search text is still displayed outside the window
image

